PR #275, opened 2022-01-28
cli: `testsys status` update

### LINKED ISSUE #265: cli: rethink testsys status
We should do some design ideation on what we want a great `testsys status` experience to be and implement it. I envision going back to a columnar table, but let's discuss.

ecpullen:
**Issue number:**
Closes #265 

**Description of changes:**
Provides a clean output for `testsys status`.

**Testing done:**
```
┌────────────────────────┬────────────┬───────────┬────────┬─────────┬────────┐
│          Name          │    Type    │   State   │ Passed │ Skipped │ Failed │
├────────────────────────┼────────────┼───────────┼────────┼─────────┼────────┤
│       Controller       │ Controller │  Running  │        │         │        │
├────────────────────────┼────────────┼───────────┼────────┼─────────┼────────┤
│      testsys-test      │  Resource  │ completed │        │         │        │
├────────────────────────┼────────────┼───────────┼────────┼─────────┼────────┤
│ testsys-test-instances │  Resource  │ completed │        │         │        │
├────────────────────────┼────────────┼───────────┼────────┼─────────┼────────┤
│       ecs-smoke        │    Test    │  passed   │   1    │    0    │   0    │
└────────────────────────┴────────────┴───────────┴────────┴─────────┴────────┘
```

**Terms of contribution:**

### DIFF
--- a/testsys/src/status.rs
+++ b/testsys/src/status.rs
@@ -12,14 +12,12 @@ use snafu::ResultExt;
 use std::collections::HashMap;
 use std::fmt::Display;
 use structopt::StructOpt;
+use tabled::{Alignment, Column, Full, MaxWidth, Modify, Style, Table, Tabled};
+use termion::terminal_size;
 
 /// Check the status of a TestSys object.
 #[derive(Debug, StructOpt)]
 pub(crate) struct Status {
-    /// Check the status of the `Test`s provided, or all `Test`s if no specific resource is provided.
-    #[structopt(long = "tests", short = "t")]
-    tests: Option<Vec<String>>,
-
     /// Check the status of the `Resource`s provided, or all `Resource`s if no specific resource is provided.
     #[structopt(long = "resources", short = "r")]
     resources: Option<Vec<String>>,
@@ -27,10 +25,6 @@ pub(crate) struct Status {
     /// Check the status of the testsys controller
     #[structopt(long, short = "c")]
     controller: bool,
-
-    /// Continue checking the status of the test/resources(s) until all have completed.
-    #[structopt(long = "wait")]
-    wait: bool,
 
     /// Output the results in JSON format.
     #[structopt(long = "json")]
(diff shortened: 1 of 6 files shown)


webern (changes requested):
We've had an out-of-band discussion and arrived at something simpler. The `--wait` use case is fuzzy and our implementation is awkward. So we can simplify things significantly

- We should eliminate the `--wait` argument and functionality.
- `testsys status` always exists 0 (unless it has encountered an error in it's execution)
- We don't need to do anything with resetting the cursor.
- We can keep the work you've done to find the terminal size and use it to format a table, all of that still leads to nice output when `--json` is not used.

ecpullen:
^ Removes the wait flag from the command and sets the command to always exit 0.

webern (approved):
😎

testsys/src/error.rs
```diff
@@ -130,4 +130,10 @@ pub(crate) enum Error {
 
     #[snafu(display("Unable to write data: {}", source))]
     Write { source: std::io::Error },
+
+    #[snafu(display("Unable to get the cursor position: {}", source))]
+    Cursor { source: std::io::Error },
+
+    #[snafu(display("Unable to get the terminal size: {}", source))]
+    TerminalSize { source: std::io::Error },
```
etungsten: Do we still need these errors variants?

testsys/tests/cli_test.rs
```diff
@@ -180,19 +122,7 @@ async fn test_system() {
...
-        "--kubeconfig",
-        cluster.kubeconfig().to_str().unwrap(),
-        "status",
-        "-r",
-        "never-destroy",
-        "--wait",
-    ])
-    .timeout(POD_TIMEOUT);
-    cmd.assert().success();
+    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
etungsten: nit: This seems extraneous if we're waiting later in the loop anyways, could wait first in the loop before doing the assertion.

testsys/src/status.rs
```diff
@@ -340,3 +298,127 @@ impl Display for ResourceResult {
...
+            min_state_width: 5,
+            min_passed_width: 6,
+            min_skipped_width: 7,
+            min_failed_width: 6,
+        }
+    }
+}
+
+impl Results {
+    fn draw(&self) {
```
etungsten, line 346: Good job figuring this out! But it almost makes me wonder if there are other crates out there we can use to do this.

etungsten (approved):

Issue #248, opened 2022-01-15
build: copy licenses


webern:
When we publish containers, we will need to copy licenses to them.

webern:
This is mostly done, but there are some TODOs in the Dockerfile.

PR #283, opened 2022-02-04
Copy licenses for rust, go, sdk tools

jpculp:
**Issue number:**

#248 

**Description of changes:**

This change generates license attribution data from our source, but also the source of external go binaries that we include. The Dockerfile was refactored to reduce the build time increases associated with this change.

**Testing done:**

`make images`

**Terms of contribution:**


Dockerfile
```diff
@@ -5,23 +5,47 @@
...
+# =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
+FROM build as build-go
+USER builder
+
+ARG GOARCH
+ARG GOROOT="/usr/libexec/go"
+
+RUN export GOARCH="${GOARCH}" ; \
+    export GOROOT="${GOROOT}" ; \
+    export PATH="${GOROOT}/bin:${PATH}"
```
webern: I don't think this actually does anything, right? Don't you need to use ENV?

Dockerfile
```diff
@@ -5,23 +5,47 @@
...
+      --spdx-data /usr/libexec/tools/spdx-data \
+      --out-dir /usr/share/licenses/testsys/vendor \
+      cargo --offline --locked Cargo.toml
 RUN --mount=type=cache,mode=0777,target=/src/target \
     cargo install --offline --locked \
       --target ${ARCH}-bottlerocket-linux-musl \
       --path . \
       --root .
 
+# TODO get licenses for boringtun
```
webern, line 47: Adding a comment for easy issue creation: add a license for boringtun

Dockerfile
```diff
@@ -101,64 +266,69 @@ FROM scratch as ecs-resource-agent
...
-    curl -fsSL "${AWS_IAM_AUTHENTICATOR_URL}" -o "${temp_dir}/${AWS_IAM_AUTHENTICATOR_URL##*/}" && \
-    chmod 0755 "${temp_dir}/${AWS_IAM_AUTHENTICATOR_URL##*/}" && \
-    mv "${temp_dir}/${AWS_IAM_AUTHENTICATOR_URL##*/}" /usr/bin/aws-iam-authenticator && \
-    rm -rf ${temp_dir}
+# Copy aws-iam-authenticator binary
+COPY --from=aws-iam-authenticator-build /tmp/aws-iam-authenticator /usr/bin/aws-iam-authenticator
+# Copy aws-iam-authenticator licenses
+COPY --from=aws-iam-authenticator-build /usr/share/licenses/aws-iam-authenticator /licenses/aws-iam-authenticator
 
+# TODO move this out, get hashes, and attribute licenses
```
webern, line 300: Commenting here for easy issue creation. aws-cli may need the same treatment as the other libraries with licenses.

Dockerfile
```diff
@@ -43,54 +67,195 @@ RUN temp_dir="$(mktemp -d --suffix wireguard-tools-setup)" && \
...
+      -o aws-iam-authenticator-${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz && \
+    grep aws-iam-authenticator-${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz \
+      /src/hashes/aws-iam-authenticator | sha512sum --check - && \
+    tar -xf aws-iam-authenticator-${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz --strip-components 1 \
+      -C aws-iam-authenticator && \
+    rm aws-iam-authenticator-${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz
+
+WORKDIR /home/builder/aws-iam-authenticator/
+RUN go mod vendor
+# TODO this might need a patch to include necessary vendored dependencies
```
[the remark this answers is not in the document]
webern: Resolving having added an issue for this.

Dockerfile
webern: #285

Dockerfile
webern: #286

webern:
^ I pushed a rebase

webern (approved):

webern:
I ran `make images` locally and it was happy.

Dockerfile
jpculp: We did this in the SDK, but I think what happens is everything that uses `FROM build-go` will inherit those exported ENVs.

hashes/kubernetes
```diff
@@ -0,0 +1,3 @@
+a53dc25aa091075d73b7f8ae9a518c019a16b821bce6cfe3ff191cf4098df8851de03ec0a9292e534045323b8300532a23e6caceae3e6dfaea00471eac16a223  k8s-1.21.6.tar.gz
+aafb8a0f3613e7511854ed972004a6c5db72fede5f1289ba14044e44c36957ff53332b0ae36762137fd32218db0435449bbb1f49a1e8d50deba8a9a7aa6f480f  kubeadm.amd64
```
jpculp: I just realized the kubeadm binaries don't have versions in the name. I'll go ahead and fix this.

Dockerfile
webern: I still don't think this has any effect. Try this Dockerfile

```dockerfile
FROM alpine as first

RUN export HELLO=world

FROM first

RUN echo $HELLO > /hello.txt

CMD ["cat", "hello.txt"]
```

`docker run -it --rm dumb-container`

Output is blank.

------

Now try 

```dockerfile
FROM alpine as first

ENV HELLO=world

FROM first

RUN echo $HELLO > /hello.txt

CMD ["cat", "hello.txt"]
```

Output is `world`

jpculp:
- replaced go exports with an ENV
- renamed some files to be consistent and show versions

Dockerfile
jpculp: Yep, seems I misunderstood. Setting path via ENV did the trick.

webern (approved):

webern:
@ecpullen and @etungsten, this gets us most of the way there for the licenses we need to include with our images before we publish them. It looks to me like this won't slow down our development cycles too much because the license stages shouldn't need to be rebuilt when we are changing the Rust sourcecode.

etungsten (approved):

Issue #288, opened 2022-02-09
build: go mod vendor is slow in some environments


webern:
In some environments `go mod vendor` in the docker builds can take a very long time. This is currently blocking #267

etungsten:
I wonder if we can try leveraging the go toolchain from the Bottlerocket SDK like we do for Bottlerocket proper, see:
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/Makefile.toml#L258-L268

The docker-go tool can be found here: https://github.com/bottlerocket-os/bottlerocket/blob/develop/tools/docker-go

webern:
> The docker-go tool can be found here

Yes. In general we seem to not have this problem for other builds including Bottlerocket, so we need to figure out what we are doing differently here.

webern:
With #290 this is no longer blocking so I will change the labels and milestone assignment.

webern:
This is addressed by #389, closing.

Issue #267, opened 2022-01-25
integrate with internal bottlerocket release process


webern:
Let's use testsys for something!

There won't be any PR associated with the, but we will close it once testsys has found its way into the Bottlerocket release process.

Tracking:
- [x] Bug Fixes
  - [x] #314
  - [x] #313
  - [x] #312
  - [x] #328
  - [x] #332
  - [x] #314
  - [x] #325
  - [x] #327
  - [x] #339
  - [x] #353  
- [x] Ergonomic Improvements
  - [x] #79
  - [x] #298
  - [x] #304 
  - [x] #324
  - [x] #323
  - [x] #326
  - [x] #329
  - [x] #331
- [ ] Support all Variants and Test Modalities
  - [x] #315 
  - [x] #316 
  - [x] #317
  - [x] #318
  - [ ] #319

webern:
Unfortunately #288 is blocking progress on this right now.

webern:
#290 unblocks progress on this.

webern:
Currently blocked by #313 

We can consider making the `sonobuoy retrieve` part optional and simply not doing it for now to unblock progress. That is, if we can't figure out the cause quickly.

webern:
We have unblocked this by eliminating the tar ball retrieving functionality for now.

mjsterckx:
Since there is only one thing left, we can use #319 to track this.

PR #290, opened 2022-02-09
build: enable docker builds on internal hosts

webern:
**Issue number:**

Provides a temporary, unblocking workaround for #288 
Enables further work on #267

**Description of changes:**

On our internal build host we need `--network=host` and `GOPROXY=direct` in order for the docker builds to work function.

After that it was discovered that `go mod vendor` for `eksctl` takes over an hour. Something is wrong here, but in order to unblock, we can skip the license inclusion of eksctl until we are ready to publish containers.

**Testing done:**

The containers build on our build host in about 10 minutes, which is reasonable starting from a clean slate.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,10 @@ USER builder
 ARG GOARCH
 ARG GOOS=linux
 ARG GOROOT="/usr/libexec/go"
+ARG GOPROXY
 
 ENV PATH="${GOROOT}/bin:${PATH}"
+ENV GOPROXY="${GOPROXY}"
 
 # =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
 FROM build as build-src
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,13 +91,16 @@ RUN mkdir eksctl && curl -L ${EKSCTL_SOURCE_URL} \
     rm eksctl_${EKSCTL_VERSION}.tar.gz
 
 WORKDIR /home/builder/eksctl/
-RUN go mod vendor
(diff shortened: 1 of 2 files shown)


webern:
^ rebase and fix a small mistake

Dockerfile
```diff
@@ -89,13 +91,16 @@ RUN mkdir eksctl && curl -L ${EKSCTL_SOURCE_URL} \
...
 WORKDIR /home/builder/eksctl/
-RUN go mod vendor
-RUN cp -p LICENSE /usr/share/licenses/eksctl && \
-    /usr/libexec/tools/bottlerocket-license-scan \
-      --clarify /src/clarify.toml \
-      --spdx-data /usr/libexec/tools/spdx-data \
-      --out-dir /usr/share/licenses/eksctl/vendor \
-      go-vendor ./vendor
+# TODO - restore this with a fix for https://github.com/bottlerocket-os/bottlerocket-test-system/issues/288
+# For reasons not yet understood, this can take an hour or more in certain environments. For now we need
```
etungsten, line 95: I wonder if we can try leveraging the go toolchain from the Bottlerocket SDK like we do for Bottlerocket proper, see:
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/Makefile.toml#L258-L268

etungsten: The `docker-go` tool can be found here: https://github.com/bottlerocket-os/bottlerocket/blob/develop/tools/docker-go

webern: Can we can add these comments to #288 and figure it out later?

etungsten (approved):

webern:
^ fix whitespace mistake

jpculp (approved):

PR #296, opened 2022-02-14
Fix glibc requirement for aws iam

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Al2 does not contain the necessary glibc version for the aws iam authenticator so we need to set `CGO_ENABLED=0` in order to use the aws iam authenticator in the sonobuoy test agent.

**Testing done:**
Sonobuoy test now performs as expected.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -206,7 +206,7 @@ RUN mkdir aws-iam-authenticator && \
 
 WORKDIR /home/builder/aws-iam-authenticator/
 RUN go mod vendor
-RUN go build -mod=vendor -o /tmp/aws-iam-authenticator \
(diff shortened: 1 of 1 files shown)


webern (approved):

etungsten (approved):

webern:
It's not obvious to me why this flag switches static vs. dynamic linking, but it looks like `0` is a good choice for release based on this: https://github.com/kubernetes-sigs/aws-iam-authenticator/blob/2cc0aa239bd03c29ded75cd58eda4e0bc76a7d64/.goreleaser.yaml#L19

Issue #292, opened 2022-02-10
Add upgrade-downgrade flag to aws-k8s, aws-ecs and vmware-k8s commands


webern:
We need to be able to run upgrade downgrade tests easily with the testsys cli.

When we give the upgrade-downgrade flag, a series of tests that depend on each other should be created (along with the resources that all of them depend on)

- initial sonobuoy check
- migrate to upgrade version
- sonobuoy upgrade version check
- migrate back to original version
- sonobuoy original version check

Some additional arguments will be required e.g.
- --starting-version
- --upgrade-version
- --tuf-repo-metadata-url
- --tuf-repo-targets-url

ecpullen:
Closed by #297

PR #297, opened 2022-02-14
cli: upgrade/downgrade testing for `aws-k8s`, `aws-ecs`, `vmware`

ecpullen:
**Issue number:**

#292 

**Description of changes:**

Adds `--upgrade-downgrade` flag to each command will extra needed information.
If an upgrade/downgrade is being run, 5 test agents are created named:
`{test_name}-initial-test`: An initial test for the respective cluster (ecs task, sonobuoy)
`{test_name}-upgrade-migration-test`: Upgrade each instance from `starting-version` to `upgrade-version`
`{test_name}-upgraded-test`: Test the instances at `upgrade-version`
`{test_name}-downgrade-migration-test`: Downgrade each instance from `upgrade-version` to `starting-version`
`{test_name}-final-test`: Test the instances at `starting-version`.

**Testing done:**
System runs successfully for `testsys run aws-ecs --upgrade-downgrade ...` and `testsys run aws-eks --upgrade-downgrade ...`.
`testsys run vmware --upgrade-downgrade` is currently untested due to a network issue.

**Terms of contribution:**

### DIFF
--- a/testsys/src/run_aws_ecs.rs
+++ b/testsys/src/run_aws_ecs.rs
@@ -1,7 +1,9 @@
 use crate::error::{self, Result};
 use bottlerocket_agents::{
-    ClusterType, Ec2Config, EcsClusterConfig, EcsTestConfig, AWS_CREDENTIALS_SECRET_NAME,
+    ClusterType, Ec2Config, EcsClusterConfig, EcsTestConfig, MigrationConfig, TufRepoConfig,
+    AWS_CREDENTIALS_SECRET_NAME,
 };
+use kube::ResourceExt;
 use kube::{api::ObjectMeta, Client};
 use maplit::btreemap;
 use model::clients::{CrdClient, ResourceClient, TestClient};
@@ -9,7 +11,9 @@ use model::constants::NAMESPACE;
 use model::{
     Agent, Configuration, DestructionPolicy, Resource, ResourceSpec, SecretName, Test, TestSpec,
 };
+use serde_json::Value;
 use snafu::ResultExt;
+use std::collections::BTreeMap;
 use structopt::StructOpt;
 
 /// Create an EKS resource, EC2 resource and run Sonobuoy.
@@ -101,6 +105,35 @@ pub(crate) struct RunAwsEcs {
     /// Name of the pull secret for the EC2 provider image.
     #[structopt(long)]
     ec2_provider_pull_secret: Option<String>,
+
+    /// Perform an upgrade downgrade test.
+    #[structopt(long, requires_all(&["starting-version", "upgrade-version"]))]
+    upgrade_downgrade: bool,
+
+    /// Starting version for an upgrade/downgrade test.
+    #[structopt(long, requires("upgrade-downgrade"))]
+    starting_version: Option<String>,
+
+    /// Version the ec2 instances should be upgraded to in an upgrade/downgrade test.
+    #[structopt(long, requires("upgrade-downgrade"))]
+    upgrade_version: Option<String>,
+
+    /// Location of the tuf repo metadata.
+    #[structopt(long, requires_all(&["tuf-repo-targets-url", "upgrade-downgrade"]))]
+    tuf_repo_metadata_url: Option<String>,
+
+    /// Location of the tuf repo targets.
+    #[structopt(long, requires_all(&["tuf-repo-metadata-url", "upgrade-downgrade"]))]
+    tuf_repo_targets_url: Option<String>,
+
+    /// Location of the migration agent image.
+    // TODO - default to an ECR public repository image
+    #[structopt(long)]
+    migration_agent_image: Option<String>,
+
+    /// Name of the pull secret for the ecs migration image (if needed).
+    #[structopt(long)]
+    migration_agent_pull_secret: Option<String>,
 }
 
 impl RunAwsEcs {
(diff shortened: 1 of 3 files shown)


testsys/src/run_aws_ecs.rs
```diff
@@ -101,6 +105,35 @@ pub(crate) struct RunAwsEcs {
...
     #[structopt(long)]
     ec2_provider_pull_secret: Option<String>,
+
+    /// Perform an upgrade downgrade test.
+    #[structopt(long, requires_all(&["starting-version", "upgrade-version"]))]
+    upgrade_downgrade: bool,
+
+    /// Starting version for an upgrade/downgrade test.
+    #[structopt(long)]
+    starting_version: Option<String>,
```
webern, line 115: Can all of these require `--upgrade-downgrade`? i.e. can we avoid accepting something like `--starting-version` when upgrade_downgrade is false?

testsys/src/run_aws_ecs.rs
```diff
@@ -111,51 +144,199 @@ impl RunAwsEcs {
...
+                )?;
+                vec![
+                    init_ecs_test,
+                    upgrade_test,
+                    upgraded_ecs_test,
+                    downgrade_test,
+                    final_ecs_test,
+                ]
+            } else {
+                return Err(error::Error::InvalidArguments{why:"If performing an upgrade/downgrade test, `starting-version`, `upgrade-version` must be provided.".to_string()});
```
webern: I'm not sure why fmt is allowing this. Can you try manually breaking this up onto a couple of lines such that the long string ends up on its own line?

testsys/src/run_aws_ecs.rs
```diff
@@ -111,51 +144,199 @@ impl RunAwsEcs {
...
+                    self.tuf_repo_targets_url.as_ref(),
+                ) {
+                    Some(TufRepoConfig {
+                        metadata_url: tuf_repo_metadata_url.to_string(),
+                        targets_url: tuf_repo_targets_url.to_string(),
+                    })
+                } else {
+                    None
+                };
+                let init_test_name = format!("{}-initial-test", self.name);
```
webern: Suggesting these names for brevity and so that they will sort in the order that they execute:

```
format!("{}-1-initial", self.name);
format!("{}-2-migrate", self.name);
format!("{}-3-migrated", self.name);
format!("{}-4-migrate", self.name);
format!("{}-5-final", self.name);
```

testsys/src/run_aws_ecs.rs
```diff
@@ -164,40 +345,52 @@ impl RunAwsEcs {
...
-                resources: vec![ec2_resource_name.clone(), cluster_resource_name.clone()],
-                depends_on: Default::default(),
+                resources: vec![
+                    ec2_resource_name.to_owned(),
+                    cluster_resource_name.to_owned(),
+                ],
+                depends_on,
                 agent: Agent {
-                    name: "ecs-test-agent".to_string(),
+                    name: format!("{}-agent", name),
```
webern: The agent name field is poorly conceived by me as a name for the program that does the work. In otherwords, this is the `ecs-test-agent`. The sonobuoy test agent is the `sonobuoy-test-agent`, etc. We should probably remove the name field from the Agent model, but for now let's keep the names generic like they were intended.

webern:
I think we should come back to this rather than trying to address it here. I think there is an opportunity to use some traits or cleverness to refactor in a way that deduplicates the code a bit more between the `vmware-k8s`, `aws-k8s` and `ecs` commands. It's tricky because of all the strongly typed entities (structopt and our own config structs). We should look into it a bit when we have a chance.

testsys/src/run_aws_ecs.rs
```diff
@@ -209,44 +402,69 @@ impl RunAwsEcs {
...
+        &self,
+        name: &str,
+        version: &str,
+        tuf_repo: Option<TufRepoConfig>,
+        secrets: Option<BTreeMap<String, SecretName>>,
+        ec2_resource_name: &str,
+        cluster_resource_name: &str,
+        depends_on: Option<Vec<String>>,
+        migration_agent_image: &str,
+        migration_agent_pull_secret: &Option<String>,
```
webern, line 434: We should try removing the options around things that are actually not optional. Below in the function body you can say `Some(tuf_repo)`. The guiding principles are: a) validate data (i.e. deal with missing data) as early as possible, and b) making function signatures that cannot be misused.

ecpullen: The tuf repo is actually optional.

testsys/src/run_aws_k8s.rs
```diff
@@ -127,28 +158,166 @@ impl RunAwsK8s {
...
+                )?;
+                vec![
+                    init_eks_test,
+                    upgrade_test,
+                    upgraded_eks_test,
+                    downgrade_test,
+                    final_eks_test,
+                ]
+            } else {
+                return Err(error::Error::InvalidArguments{why:"If performing an upgrade/downgrade test, `starting-version`, `upgrade-version` must be provided.".to_string()});
```
webern: The same fix for wherever these super long lines are would be nice.

webern (approved):

webern:
There is now a commit named `Test naming change` that I don't think is intended to be there.

testsys/src/run_aws_ecs.rs
```diff
@@ -111,51 +144,203 @@ impl RunAwsEcs {
...
+                        metadata_url: tuf_repo_metadata_url.to_string(),
+                        targets_url: tuf_repo_targets_url.to_string(),
+                    })
+                } else {
+                    None
+                };
+                let init_test_name = format!("{}-1-initial", self.name);
+                let upgrade_test_name = format!("{}-2-migrate", self.name);
+                let upgraded_test_name = format!("{}-3-migrated", self.name);
+                let downgrade_test_name = format!("{}-4-migrate", self.name);
```
etungsten, line 181: should we include the versions in the test name?
```suggestion
                let upgrade_test_name = format!("{}-2-migrate-to-{}-from-{}", self.name, upgrade_version, starting_version);
                let upgraded_test_name = format!("{}-3-test-after-migrate", self.name);
                let downgrade_test_name = format!("{}-4-migrate-to-{}-from-{}", self.name, starting_version, upgrade_version);
```

testsys/src/run_aws_ecs.rs
```diff
@@ -209,44 +406,69 @@ impl RunAwsEcs {
...
+            },
+            spec: TestSpec {
+                resources: vec![
+                    ec2_resource_name.to_owned(),
+                    cluster_resource_name.to_owned(),
+                ],
+                depends_on,
+                agent: Agent {
+                    name: "ecs-test-agent".to_string(),
+                    image: migration_agent_image.to_string(),
```
etungsten, line 462: Is the agent name accurate? Would this be clearer?
```suggestion
                    name: "ecs-migration-test-agent".to_string(),
                    image: migration_agent_image.to_string(),
```

etungsten (approved):

testsys/src/run_aws_ecs.rs
webern: Maybe. Note that, if we do, I think dots are not allowed.

testsys/src/run_aws_ecs.rs
webern: The name here should basically match the name of the crate that produced the image. I commented on this earlier. The `name` field of the `Agent` model is poorly conceived. I thought of it as a name for the program, e.g.
- ecs-test-agent
- sonobuoy-test-agent
- migration-test-agent
- eks-resource-agent

I don't think we are using it for anything at all and we should probably remove it from the model since it's a source of confusion that adds no value.

webern: Created #306

PR #301, opened 2022-02-15
resource-agent: enable keep-running

### LINKED ISSUE #79: agents: stay running after test or error
As written the test_agent will exit if the `Runner` returns an error.

https://github.com/bottlerocket-os/bottlerocket-test-system/blob/6e61ed8a6a219aeed5522f572f2257394fe2b18d/test-agent/src/agent.rs#L74

Instead it should send the error to k8s but leave the program running until the test controller tells it to terminate. The use case is that you may want to inspect things before terminate cleans up and the container goes down.

Edit: Actually, a user might want to keep the pod running even if the test succeeds, and we anticipate making it possible to copy data out of the pod. So we should have some sort of flag in the CRD, e.g. `keep_running: true`, and loop checking this after `terminate`. If it becomes false, we proceed to exit the program, otherwise we keep looping.

webern:
**Issue number:**

Closes #79
Needed for troubleshooting during #267

**Description of changes:**

- In the resource-agent, make sure we stay running when asked to.
- In the test-agent, print error messages before going into our wait loop.
- In the CLI, make these available to the aws-k8s, aws-ecs and vmware-k8s commands.
- Fix the logging init function.

**Testing done:**

Tested using `testsys run aws-k8s` and it looks good.

**Terms of contribution:**

### DIFF
--- a/bottlerocket-agents/src/lib.rs
+++ b/bottlerocket-agents/src/lib.rs
@@ -368,11 +368,11 @@ pub fn init_agent_logger(bin_crate: &str, log_level: Option<LevelFilter>) {
                 .filter_level(LevelFilter::Error)
                 // Set all of our crates to the desired level.
                 .filter(Some(bin_crate), log_level)
-                .filter(Some("agent-common"), log_level)
-                .filter(Some("bottlerocket-agents"), log_level)
+                .filter(Some("agent_common"), log_level)
+                .filter(Some("bottlerocket_agents"), log_level)
                 .filter(Some("model"), log_level)
-                .filter(Some("resource-agent"), log_level)
-                .filter(Some("test-agent"), log_level)
+                .filter(Some("resource_agent"), log_level)
+                .filter(Some("test_agent"), log_level)
                 .init();
         }
     }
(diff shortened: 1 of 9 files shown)


webern:
^ rebase

webern:
^ rebase

webern:
^ rebase

webern:
^ rebase

webern:
^ rebase

webern:
^ fix logger init

bottlerocket-agents/src/lib.rs
```diff
@@ -368,11 +368,11 @@ pub fn init_agent_logger(bin_crate: &str, log_level: Option<LevelFilter>) {
                 .filter_level(LevelFilter::Error)
                 // Set all of our crates to the desired level.
                 .filter(Some(bin_crate), log_level)
-                .filter(Some("agent-common"), log_level)
+                .filter(Some("agent_common"), log_level)
                 .filter(Some("bottlerocket-agents"), log_level)
```
etungsten: This doesn't have to be `bottlerocket_agents`?

etungsten (approved):

webern:
^ `bottlerocket-agents` -> `bottlerocket_agents`

ecpullen (approved):

PR #305, opened 2022-02-17
eks-provider: improve instance profile search

webern:
Make the search for an instance profile deterministic.

**Issue number:**

A fix for a problem encountered during #267

**Description of changes:**

Our assumptions about instance profile naming and role naming were not correct. And in any event, it is possible to do a deterministic search for the exact role name that we care about.

**Testing done:**

It worked! We saw the eks-provider succeed in the environment where it was previously failing.

**Terms of contribution:**

### DIFF
--- a/bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -437,7 +437,7 @@ async fn created_cluster(
 
     let node_instance_role = cluster_iam_identity_mapping(cluster_name, region)?;
     let iam_instance_profile_arn =
-        instance_profile(&aws_clients.iam_client, cluster_name, &node_instance_role).await?;
+        instance_profile_arn(&aws_clients.iam_client, &node_instance_role).await?;
 
     Ok(CreatedCluster {
         cluster_name: cluster_name.to_string(),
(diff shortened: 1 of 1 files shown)


webern:
^ clippy is smart

webern:
^ rebase

webern:
Ugh, this implementation is wrong because I thought I had the role name, but instead we have the role arn. I need to convert the role arn to a role name before running the logic:

```
Provider error: An error left resources behind that can be destroyed, Unable to find an instance profile for role named
'arn:aws:iam::359404537045:role/eksctl-testsys-nodegroup-testsys-NodeInstanceRole-1F52WG29KMPW6'
```

webern:
^ rebase

webern:
^ use list_instance_profiles_for_role

webern:
^ rebase

etungsten (dismissed):

bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -644,53 +644,73 @@ async fn security_group(
...
+    if instance_profiles.len() > 1 {
+        return Err(ProviderError::new_with_context(
+            Resources::Remaining,
+            format!(
+                "More than one instance profile was found for role name \
+                '{}' which was derived from arn '{}'",
+                role_name, role_arn
+            ),
+        ));
+    }
```
etungsten: I don't think this should be a hard error, and instead we can just take the first associated instance profile and try running with it.
nit: the rolename is substring of role_arn.
```suggestion
    if instance_profiles.len() > 1 {
        return Err(ProviderError::new_with_context(
            Resources::Remaining,
            format!(
                "More than one instance profile was found for role '{}',
                role_arn
            ),
        ));
    }
```
applies elsewhere too.

bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -644,53 +644,73 @@ async fn security_group(
...
+
+    let instance_profile = instance_profiles.into_iter().next().ok_or_else(|| {
+        ProviderError::new_with_context(
+            Resources::Remaining,
+            format!(
+                "More than one instance profile was found for role name \
+                '{}' which was derived from arn '{}'",
+                role_name, role_arn
+            ),
```
etungsten: This should mean there's no instance profile for the role at all.

```suggestion
    let instance_profile = instance_profiles.into_iter().next().ok_or_else(|| {
        ProviderError::new_with_context(
            Resources::Remaining,
            format!(
                "No instance profile was found for role '{}', role_arn
            ),
```

ecpullen (approved):

bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
webern: Good idea regarding simplifying the error messages.

Regarding raising an error if more than one instance profile is returned...

So, hopefully it is academic and is actually impossible by some constraint of the AWS API (this part of the model is weird). But, if it did occur, I think it would be better for an error to be raised notifying the user of this problem. Otherwise they would have some extremely difficult behavior where the agent doesn't have the right permissions and it could possibly even be intermittent. It seems like something that could take hours or days to figure out.

bottlerocket-agents/src/bin/eks-resource-agent/eks_provider.rs
webern: Good catch.

webern:
^ simplify error messages

webern:
^ fix a git mistake

etungsten (approved):

PR #333, opened 2022-02-27
improve sonobuoy testing

### LINKED ISSUE #324: testsys delete --include-resources should be ok if a resource is missing
If a dependent resource does not exist during `--include-resources`, this should not be an error. The command should print a statement saying the resource does not exist, then go on to it's next task.

### LINKED ISSUE #328: sometimes pods take forever to terminate
Deleting a test can take a long time when the test agent pod stays in the Terminating state for a while. Perhaps we can tell kubernetes to use shorter grace period when terminating test-agent and resource-agent pods.

Edit: perhaps we need to listen for the graceful shutdown signal, whether its a SIGINT or whatever it is and exit.

### LINKED ISSUE #331: when sonobuoy tests fail we need to list failed tests
When sonobuoy tests fail, it is a lot of work to go get the list of failed tests. We should at least log these.

### LINKED ISSUE #332: use private subnet by default for ec2 instances
At least some of the time, when providing EKS nodes, the EC2 provider puts them in a public subnet. This is problematic because having a public IP address changes Sonobuoy behavior in a way that causes certain tests to fail.

webern:
**Issue number:**

Closes #332
Attempts to improve #328
Closes #324
Closes #331

**Description of changes:**

```
When the testsys CLI is creating EC2 instances for EKS and ECS
clusters, it should default to using private subnets. Later if we need
public subnets for some reason, we can make this configurable.
```

```
The background propagation policy is described as "Allow[ing] the
garbage collector to delete the dependents in the background.
Hopefully this will speed up deletes.
```

```
When the CLI is performing a delete, and the item to be deleted does
not exist, we should not error out. Instead we should inform the user
that the item was not found.
```

```
sonobuoy-test-agent: print results in logs

This will make it easier to find out what tests failed. Sometimes all
you need is the name. At least now we can look at the pod logs.
```

**Testing done:**

- using private subnets works
- the background propagation policy... hard to say if it had any effect, but no harm was done
- print results in logs: this failed because there was a --kubeconfig argument. i have removed that argument since seeing that problem. the good news is, it didn't crash the agent since the code i wrote allows sonobuoy results to fail... so no harm done and i think it will work now
- used `testsys delete` to delete a test where the resources were missing and saw:

```
Resource 'aarch64-aws-k8s-121' does not exist. Skipping.
Resource 'aarch64-aws-k8s-121-instances' does not exist. Skipping.
Deleting test 'conformance-aarch64-aws-k8s-121-test'
Test 'conformance-aarch64-aws-k8s-121-test' has been deleted
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket-agents/src/sonobuoy.rs
+++ b/bottlerocket-agents/src/sonobuoy.rs
@@ -1,5 +1,5 @@
-use crate::{error, SonobuoyConfig};
-use log::{info, trace};
+use crate::{error, SonobuoyConfig, SONOBUOY_RESULTS_FILENAME};
+use log::{error, info, trace};
 use model::{Outcome, TestResults};
 use serde::{Deserialize, Serialize};
 use serde_plain::{derive_display_from_serialize, derive_fromstr_from_deserialize};
@@ -74,13 +74,29 @@ pub async fn run_sonobuoy(
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
     info!("Running sonobuoy retrieve");
+    let results_filepath = results_dir.join(SONOBUOY_RESULTS_FILENAME);
     let status = Command::new("/usr/bin/sonobuoy")
-        .current_dir(results_dir.to_str().context(error::ResultsLocationSnafu)?)
         .args(kubeconfig_arg.to_owned())
         .arg("retrieve")
+        .arg("--filename")
+        .arg(results_filepath.as_os_str())
         .status()
         .context(error::SonobuoyProcessSnafu)?;
     ensure!(status.success(), error::SonobuoyRunSnafu);
+
+    info!("Sonobuoy testing has completed, printing results");
+    let sonobuoy_results_exist_status = Command::new("/usr/bin/sonobuoy")
+        .arg("results")
+        .arg(results_filepath.as_os_str())
+        .status()
+        .context(error::SonobuoyProcessSnafu)?;
+
+    if !sonobuoy_results_exist_status.success() {
+        error!(
+            "Bad exit code from 'sonobuoy results': exit {}",
+            sonobuoy_results_exist_status.code().unwrap_or(1)
+        )
+    }
 
     info!("Getting Sonobuoy status");
     let run_result = Command::new("/usr/bin/sonobuoy")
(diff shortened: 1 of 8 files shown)


webern:
^ remove resource destruction work from this PR

webern:
^ Compromise with clippy on what I was trying to do. Rename the trait that I defined, hopefully it reads well.

webern:
^ try to improve the 'is found' trait some more.

webern:
^ Fix a comment

webern:
^ remove errant kubeconfig arg in `sonobuoy results` command.

webern:
^ when deleting, also prevent an error if the GET operation fails because an item does not exist.

bottlerocket-agents/src/sonobuoy.rs
```diff
@@ -74,14 +74,30 @@ pub async fn run_sonobuoy(
...
-        .current_dir(results_dir.to_str().context(error::ResultsLocationSnafu)?)
         .args(kubeconfig_arg.to_owned())
         .arg("retrieve")
+        .arg("--filename")
+        .arg(results_filepath.as_os_str())
         .status()
         .context(error::SonobuoyProcessSnafu)?;
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
+    info!("Sonobuoy testing has completed, printing results");
```
ecpullen, line 87: Where are we printing the results?

controller/src/job/mod.rs
```diff
@@ -109,7 +109,7 @@ pub(crate) async fn delete_job(k8s_client: kube::Client, name: &str) -> JobResul
             &DeleteParams {
                 dry_run: false,
                 grace_period_seconds: Some(0),
-                propagation_policy: Some(PropagationPolicy::Foreground),
+                propagation_policy: Some(PropagationPolicy::Background),
```
ecpullen, line 112: Does this mean that the `Resource` will be deleted as soon as the delete starts? If so, it may become problematic. The ecs deletions flow require a deletion to be fully completed before starting the next step.

webern: No this is for deletion of the Job (i.e. the Pod). So once a test agent or resource agent is already running, and we want to delete that job (i.e. after it's work is done and we are just terminating it), we are allowing that pod termination to happen in the background.

bottlerocket-agents/src/sonobuoy.rs
webern: It goes to the logs (stdout) because I am using Cmd.status() and not Cmd.output()

ecpullen (approved):

etungsten (approved):

PR #334, opened 2022-02-28
optionally ignore resource destruction error

### LINKED ISSUE #323: provide a way to opt out of resource deletion after error
If a resource destruction pod results in an error, we have left the user in a bad state. It is very hard to get the Resource deleted once this happens. We need to provide a way for the user to take responsibility for deletion of the "physical" resources and to say to testsys, "I will take care of the problem, delete the Resource object --force".

webern:
**Issue number:**

Thoughts on #323

**Description of changes:**

An idea for abandoning resources when the destruction pod fails

**Testing done:**

**Terms of contribution:**


webern:
closing this for now as it wasn't really a complete thought and we have other stuff to work on first

PR #342, opened 2022-03-03
cli: `restart test`

### LINKED ISSUE #340: cli: restart failed tests
If a test fails due to, e.g. AWS SDK throttling... or for any other reason where we don't care about preserving the state of the failed test run, the CLI could "restart" the test by deleting and recreating the Test CRD (without deleting the resources)

ecpullen:
**Issue number:**

Closes #340 

**Description of changes:**

Adds `testsys restart test <test-name>`. This command deletes the test object and then re-creates it.

**Testing done:**

Ran `testsys run aws-k8s ...` and after test completion used `testsys restart test ...` and the sonobuoy test was deleted and a new one was started.

**Terms of contribution:**

### DIFF
--- a/testsys/src/restart_test.rs
+++ b/testsys/src/restart_test.rs
@@ -1,0 +1,40 @@
+use crate::error::{self, Result};
+use kube::Client;
+use model::clients::{CrdClient, TestClient};
+use snafu::ResultExt;
+use structopt::StructOpt;
+
+/// Restart an object from a testsys cluster.
+#[derive(Debug, StructOpt)]
+pub(crate) struct RestartTest {
+    /// The name of the test to be restarted.
+    #[structopt()]
+    test_name: String,
+}
+
+impl RestartTest {
+    pub(crate) async fn run(&self, k8s_client: Client) -> Result<()> {
+        let test_client = TestClient::new_from_k8s_client(k8s_client.clone());
+        let mut test = test_client
+            .get(&self.test_name)
+            .await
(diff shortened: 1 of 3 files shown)


testsys/src/restart_test.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub(crate) async fn run(&self, k8s_client: Client) -> Result<()> {
+        let test_client = TestClient::new_from_k8s_client(k8s_client.clone());
+        let mut test = test_client
+            .get(&self.test_name)
+            .await
+            .context(error::GetSnafu {
+                what: self.test_name.clone(),
+            })?;
+        // Created objects are not allowed to have `resource_version` set.
+        test.metadata.resource_version = None;
```
webern, line 25: You also need to set `test.status` to `None`.

webern (changes requested):
Nice, that was fast. We need to obliterate the `status` otherwise I think the controller wouldn't see this as a new test.

testsys/src/restart_test.rs
ecpullen: The status is automatically reset. We can do it explicitly if you'd prefer.

webern: Oh. Well, I guess it would make me feel better to see it even if it is not strictly necessary.

ecpullen:
^ reset `test.status`

webern (approved):

etungsten (approved):

Issue #46, opened 2021-07-07
cli: config file


webern:
It may be convenient or necessary to store config parameters in a config file. If so, design one.

- depends on #44
- depends on #45
- depends on #38
- depends on #33

`id: cli_config_file`

PR #348, opened 2022-03-09
re-organize bottlerocket-specific things

webern:
**Issue number:**

N/A but possibly a precursor to #46

**Description of changes:**

It may be hard to follow the diff but basically we are creating three directories/crates. (Note, however, that we will not move `testsys` into the Bottlerocket directory currently because there are too many PRs going on involving it. Once there are no PRs open on the testsys CLI, I will move that into the directory as shown below).

```
bottlerocket/agents
bottlerocket/testsys
bottlerocket/types
```

`bottlerocket/types` is for common code, like the configs used by both the CLI and the agents.
Moving the CLI into a `bottlerocket` directory should make it more clear that this is for tight integration with Bottlerocket. Basically the `bottlerocket` directory are the parts that could end up being in the Bottlerocket mono-repo (or else interdependent with it)

**Testing done:**

None just yet, but the compiler is happy.

**Terms of contribution:**


webern:
^ fixes

ecpullen (approved):

etungsten (approved):

PR #351, opened 2022-03-14
Eks single describe cluster call

### LINKED ISSUE #339: eks-provider: make fewer describe cluster calls
To help with #325, we should try to make only one call to eks describe-cluster during the eks agent's `create` workflow. The first time we do this is for "does_cluster_exist". After that we should hold an Option where we have Some if the cluster did exist. We could do some small wrapper around the client that caches describe-describe cluster.

ecpullen:
**Issue number:**

Closes #339 

**Description of changes:**

Extracts a `Cluster` from a single `describe_cluster` call to reduce the number of api calls.

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -1,7 +1,7 @@
 use aws_sdk_ec2::model::{Filter, SecurityGroup, Subnet};
 use aws_sdk_ec2::types::SdkError;
 use aws_sdk_eks::error::{DescribeClusterError, DescribeClusterErrorKind};
-use aws_sdk_eks::model::IpFamily;
+use aws_sdk_eks::model::{Cluster, IpFamily};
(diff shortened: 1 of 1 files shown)


webern (commented):
In line 165 (can't add comments for unchanged lines) we call `is_cluster_creation_required` which uses `describe_cluster` to determine whether or not the cluster already exists. I'm trying to think of a way that we can use the describe cluster result from that function to avoid calling again later in `created_cluster` in the case that the EKS cluster already existed.

webern (approved):
> In line 165 (can't add comments for unchanged lines) we call `is_cluster_creation_required` which uses `describe_cluster` to determine whether or not the cluster already exists. I'm trying to think of a way that we can use the describe cluster result from that function to avoid calling again later in `created_cluster` in the case that the EKS cluster already existed.

Since we've eliminated so many calls and we're left with only two, and since it gets ugly to eliminate the "does_cluster_exist" call, then I think we can call this good. Nice job.

etungsten (approved):

PR #363, opened 2022-03-30
move cli

webern:
**Issue number:**

N/A

**Description of changes:**

This moves the CLI formerly known as `testsys` to the `bottlerocket` directory. It is renamed to avoid confusion with the CLI that we will add in Bottlerocket.

This PR is built on #362 so some extra diff will be removed upon merging #362.

**Terms of contribution:**


webern:
I'm going to close this and come back to it later.

PR #384, opened 2022-04-27
TestManager

### LINKED ISSUE #367: cli: make testsys into a library
Make almost everything about the testsys CLI available in a lib.rs file so that a clone of it can easily be created in the Bottlerocket repo.

ecpullen:
**Issue number:**

Closed #367 

**Description of changes:**

There's a lot...

The `TestManager` provides operations for coordinating the creation, operation, deletion and observation of `Test` and `Resource` objects. It understands the dependencies between `Test`s and `Resource`s.
The new `testsys` cli demonstrates the functionality of the new client.

**Testing done:**

**Terms of contribution:**


model/Cargo.toml
```diff
@@ -21,8 +22,13 @@ schemars = "0"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 serde_plain = "1"
+serde_yaml = "0.8"
 snafu = "0.7"
-tokio =  { version = "1", features = ["rt-multi-thread"] }
+tabled = "0.6"
+tokio =  { version = "1", features = ["rt-multi-thread", "sync", "fs"] }
+tokio-util = "0.7"
+async-stream =  { version = "0.3"}
```
webern: sort the list

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+use crate::{
+    clients::AllowNotFound,
+    constants::{LABEL_COMPONENT, NAMESPACE, TESTSYS_RESULTS_FILE},
+    system::{
+        agent_cluster_role, agent_cluster_role_binding, agent_service_account,
+        controller_cluster_role, controller_cluster_role_binding, controller_deployment,
+        controller_service_account, testsys_namespace, AgentType,
+    },
+    Resource, SecretName, TaskState, Test,
+};
```
webern: Sorry for the nit, but to match the style in all our projects, we shouldn't group things like this, instead it should be:

```rust
use crate::clients::AllowNotFound
use crate::constants::{LABEL_COMPONENT, NAMESPACE, TESTSYS_RESULTS_FILE}
```

etc.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+    path::{Path, PathBuf},
+    thread,
+    time::Duration,
+};
+use std::{collections::HashMap, fmt::Debug};
+use tabled::{Alignment, Full, MaxWidth, MinWidth, Modify, Style, Table, Tabled};
+use tokio::io::AsyncWriteExt;
+use topological_sort::TopologicalSort;
+
+/// TestsysClient should be used as the entrypoint for any user interaction with `bottlerocket-test-system`
```
webern: This doc comment should describe the purpose of the struct. I don't think it's true that this is the only entrypoint for any user interaction. If that were the case then we could make TestClient and ResourceClient private. Instead I think the purpose of this struct is to provide an API abstraction which simplifies common interactions.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+        })?;
+        test_client.wait_for_deletion(name).await;
+        test_client.create(test).await.context(ClientSnafu {
+            action: "create new test",
+        })?;
+        Ok(())
+    }
+
+    /// Takes a path to a yaml manifest and creates a set of `Crd`s through deserialization
+    pub fn read_manifest(path: &Path) -> Result<Vec<Crd>> {
```
webern: Let's move this out of the struct and make it a free-standing function since it doesn't use `self`. Also this should be more descriptive about what a "yaml manifest" is.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+    pub fn test_client(&self) -> TestClient {
+        TestClient::new_from_k8s_client(self.k8s_client.clone())
+    }
+
+    /// Create a `ResourceClient`
+    pub fn resource_client(&self) -> ResourceClient {
+        ResourceClient::new_from_k8s_client(self.k8s_client.clone())
+    }
+
+    pub async fn create_image_secret(
```
webern:
```suggestion
    pub async fn create_image_pull_secret(
```

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+        &self,
+        name: &str,
+        username: &str,
+        password: &str,
+        image_url: &str,
+    ) -> Result<()> {
+        // Create secret for controller image pull.
```
webern:
```suggestion
        // Create docker config json for the image pull secret.
```

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+        let (image, secret) = match controller_config {
+            ImageConfig::WithCreds { secret, image } => (image, Some(secret)),
+            ImageConfig::Image(image) => (image, None),
+        };
+        self.create_deployment(image, secret).await?;
+
+        Ok(())
+    }
+
+    /// Restart a crd object
```
webern: Describe exactly what this does. "Restart a CRD object" doesn't really mean anything.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+            })?;
+            let crd: Crd = serde_yaml::from_value(value).context(SerdeYamlSnafu {
+                action: "deserialize manifest",
+            })?;
+            crds.push(crd);
+        }
+        Ok(crds)
+    }
+
+    pub async fn create_object(&self, crd: Crd) -> Result<()> {
```
webern: Normally `create` will return the created object in an API.

Also, I don't think all three of these methods are needed.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+        Some(count) => format!("{}", count),
+        None => "".to_string(),
+    }
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// The error type for `TestsysClient`
+#[derive(Debug, Snafu)]
+pub enum Error {
```
webern: ~~I'm not sure if you should experience a different Error type when using the `TestsysClient`. I think this should be wrapped into a `crate::Error` for pub functions. Also this is another opportunity for a file breakout, `error.rs` for the module.~~

Edit: We already do this where we have a public `client:Error`, and I just realized that we could have a circle of errors so... resolving.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+            objects = self.add_dependencies_to_vec(objects).await?;
+        }
+        Ok(self.delete_sorted_resources(Self::vec_to_deletion_order(objects)))
+    }
+
+    /// Delete the objects after a failed deletion attempt.
+    /// Warning: the resources created must be manually removed.
+    /// /// The finalizers will be removed from the resource and the resource will be deleted.
+    /// The job for resource deletion will also be deleted.
+    /// This should only be used if a resource has already failed to delete.
```
webern:
```suggestion
    /// Delete the objects after a failed deletion attempt.
    /// Warning: the physical resources will not be delete.
    /// The finalizers will be removed from the resource and the resource will be deleted.
    /// The k8s job for resource deletion will also be deleted.
    /// This should only be used if a resource has already failed to delete.
```

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+    pub async fn force_delete(&self, selection_params: &SelectionParams) -> Result<()> {
+        let objects = self.list(selection_params).await?;
+        for object in objects {
+            self.force_delete_object(object.into()).await?
+        }
+        Ok(())
+    }
+
+    pub async fn status(
```
webern: Document the `status` function.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+        };
+        let crds = self.list(selection_params).await?;
+
+        Ok(Status {
+            controller_status,
+            crds,
+        })
+    }
+
+    pub async fn logs(
```
webern: Document the `logs` functions

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+            objects = self.add_dependencies_to_vec(objects).await?;
+        }
+        self.stream_logs(
+            objects.into_iter().map(|object| object.into()).collect(),
+            follow,
+        )
+        .await
+    }
+
+    pub async fn write_results(&self, test_name: &str, destination: &Path) -> Result<()> {
```
webern: This needs to be documented and it might need a better name.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+                        .items,
+                );
+                pods
+            }
+        })
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum CrdName {
```
webern: I'm looking for ways to reduce this Rust file's size. I think `Crd` and `CrdName` could go to its own file. Also it should be clearly documented what it is.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+}
+
+#[derive(Debug)]
+pub enum DeleteEvent {
+    Starting(CrdName),
+    Deleted(CrdName),
+    Failed(CrdName),
+}
+
+pub enum SelectionParams {
```
webern: Possibly `SelectionParams` and `Status` and `ResultsRow` could go in their own file.

I'm not actually sure that all of this belongs in the `clients` module. This collection of files might be its own module directory as a sibling to `clients`.

model/src/clients/testsys_client.rs
```diff
@@ -0,0 +1,1301 @@
...
+
+    #[snafu(display("Unable to {}: {}", action, source))]
+    SerdeYaml {
+        action: String,
+        source: serde_yaml::Error,
+    },
+}
+
+#[derive(Serialize)]
+pub(crate) struct DockerConfigJson {
```
webern: Small types like this I might suggest putting in the `mod.rs` (not `client/mod.rs` though, I think there needs to be a sibling directory, e.g. `utils/mod.rs`)

webern (commented):
I'm going to save some of the feedback I have so far.

One big thing is that I don't see this as a "Client", at least not in the sense that we have used the words in `TestClient` and `ResourceClient`.

I think the `TestsysClient` is actually a `Utils` object, and I think it should be in a directory that is a sibling to `clients`. e.g. `src/utils/mod.rs`. Then the huge file could be broken up into some logical units within that directory.

model/src/test_manager/manager.rs
```diff
@@ -0,0 +1,585 @@
...
+        let (image, secret) = match controller_config {
+            ImageConfig::WithCreds { secret, image } => (image, Some(secret)),
+            ImageConfig::Image(image) => (image, None),
+        };
+        self.create_deployment(image, secret).await?;
+
+        Ok(())
+    }
+
+    /// Restart a crd object by deleting the crd from the cluster and adding a copy of it with its status cleared.
```
webern, line 152:
```suggestion
    /// Restart a crd object by deleting the crd from the cluster and adding a copy of it with its
    /// status cleared.
```

webern (commented):
I like the changes. One nit that goes throughout. Please set a ruler at 100 in your IDE and wrap comments so they do not extend past the ruler. I'm not sure what you are using as your comment-wrapping guide but 100 is what we use in all our projects.

model/src/test_manager/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+            CrdName::Test(name) => name,
+            CrdName::Resource(name) => name,
+        }
+    }
+}
+
+/// `Crd` provides an interface to combine `Test` and `Resource` when actions can be performed on both.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum Crd {
```
webern: `Crd` and `CrdName` seem very primitive and associated with the underlying types themselves. Perhaps the best place for these is actually `src/lib.rs`.  I could see them being used in `crd_ext.rs`, if that ever happens then that would probably be the right file.

model/src/test_manager/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+mod delete;
+mod error;
+mod install;
+mod logs;
+mod manager;
+mod status;
+
+/// `CrdName` provides a way of determing which type of testsys object a name refers to.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum CrdName {
```
webern: Should we consider?:

```rust
pub enum CrdType {
    Test,
    Resource,
}

pub struct CrdName {
    name: Cow<str>,
    crd_type: CrdType,
}
```

model/src/test_manager/manager.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    // =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
+
+    /// Retry attempts for creating or updating an object.
+    const MAX_RETRIES: i32 = 3;
+    /// Timeout for object creation/update retries.
+    const BACKOFF_MS: u64 = 500;
+
+    /// Create or update an existing k8s object
+    pub(super) async fn create_or_update<T>(
```
webern: For such a large object, I think it would clean things up further to create a file `manager_impl.rs` and move non-`pub` functions and constants there. That way, when scrolling through this file everything would be `pub`.

model/src/test_manager/status.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+fn display_option(o: &Option<u64>) -> String {
+    match o {
+        Some(count) => format!("{}", count),
+        None => "".to_string(),
+    }
+}
+
+impl Crd {
```
webern: Having this `impl Crd` over here doing logic on behalf of `ResultRow` seems off. I believe this will work:

```rust
impl From<&Crd> for Vec<ResultRow> {
    fn from(crd: &Crd) -> Self {
        Vec::new()
    } 
}
```

I was surprised by this, I guess you are allowed to impl traits for Vec<T>. I think this is explained in *Rust for Rustaceans*.

This works:

```rust
struct Foo{}

struct Bar{}

impl From<&Foo> for Vec<Bar> {
    fn from(_foo: &Foo) -> Self {
        Vec::new()
    }
}
```

This does not work:

```rust
struct Foo{}

impl Vec<Foo> {
    fn bar() {}
}
```

model/src/test_manager/status.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl Status {
+    pub(super) fn new(controller_status: Option<PodStatus>, crds: Vec<Crd>) -> Self {
+        Self {
+            controller_status,
+            crds,
+        }
+    }
+
+    fn create_table(&self) -> Table {
```
webern: This should be either:

```rust
impl From<AsRef<Status>> for Table
```

Or:

```rust
impl Table {
  pub fn new(status: &Status) -> Self
}
```

model/src/test_manager/status.rs
```diff
@@ -0,0 +1,145 @@
...
+        }
+        results.sort_by(|a, b| a.name.cmp(&b.name));
+
+        Table::new(results)
+            .with(Style::blank())
+            .with(Modify::new(Full).with(Alignment::left()))
+    }
+
+    /// Create a table containing the status of all objects.
+    pub fn to_string(&self, width: usize) -> String {
```
webern: This could be

`pub fn write_table<W>(&self, write: &mut W, width: usize)`

Then if we have a `DEFAULT_WIDTH`:

```rust
impl Display for Table {
    fn fmt(&self) -> Result<()> {
        let mut some_cursor: CursorThing<String>::new();
        self.write(&mut some_cursor, DEFAULT_WIDTH)?;
        Ok(some_cursor.to_string())
    }
}
```

testsys/src/add_secret.rs
```diff
@@ -0,0 +1,27 @@
...
+    #[clap(subcommand)]
+    command: Command,
+}
+
+#[derive(Debug, Parser)]
+enum Command {
+    /// Create a secret for image pulls.
+    Image(add_secret_image::AddSecretImage),
+    /// Create a secret from key value pairs.
+    Map(add_secret_map::AddSecretMap),
```
webern: Here's a case where I think we should put these things together into one file. Let's pull AddSecretImage and AddSecretMap into this file.

testsys/src/error.rs
```diff
@@ -0,0 +1,96 @@
+use std::fmt::{Display, Formatter};
```
webern: I believe this entire file can go away if you use `anyhow::{Error, Result}`.

testsys/src/logs.rs
```diff
@@ -0,0 +1,48 @@
...
+    #[clap(long, short)]
+    follow: bool,
+}
+
+impl Logs {
+    pub(crate) async fn run(self, client: TestManager) -> Result<()> {
+        let crd = match (self.test, self.resource) {
+            (Some(test), None) => CrdName::Test(test),
+            (None, Some(resource)) => CrdName::Resource(resource),
+            _ => return Err(Error::new_with_context("Invalid arguments were provided. Exactly on of `--test` and `--resource` must be used.")),
```
webern:
```suggestion
            _ => return Err(Error::new_with_context("Invalid arguments were provided. Exactly one of `--test` and `--resource` must be used.")),
```

model/src/test_manager/logs.rs
```diff
@@ -0,0 +1,164 @@
...
+use futures::{SinkExt, Stream, StreamExt};
+use k8s_openapi::{api::core::v1::Pod, chrono::NaiveDateTime};
+use kube::{api::LogParams, Api, ResourceExt};
+use regex::Regex;
+use snafu::ResultExt;
+use std::{cmp::Ordering, thread, time::Duration};
+
+impl TestManager {
+    /// Create a single ordered stream with logs from all `objects`.
+    pub(super) async fn stream_logs(
```
webern: This is impressive, but very complicated and solves a use case that might not exist. I'm not sure that I would ever want the logs from multiple pods to be interleaved like this.

Instead, I think we should provide the logs for a single pod only, and make it very easy to identify the desired pod.

Basically I think the functionality needed here is, given a test name, return the pod name. Given a resource name and phase (creation or destruction) return the pod name. Then another function that streams logs given a pod name (basically just wrapping and hiding kuber-rs).

The CLI can then expose ergonomics like `testsys logs --test foo` and `testsys logs --resource bar --creation`

testsys/src/install.rs
```diff
@@ -0,0 +1,33 @@
...
+/// The install subcommand is responsible for putting all of the necessary components for testsys in
+/// a k8s cluster.
+#[derive(Debug, Parser)]
+pub(crate) struct Install {
+    /// Controller image pull secret
+    #[clap(long = "controller-image-secret", short = 's')]
+    secret: Option<String>,
+
+    /// Controller image uri
+    // Todo add default controller_uri after images are published.
```
webern: nit: follows the TODO syntax elsewhere and that IDEs tend to recognize

```suggestion
    // TODO - add default controller_uri after images are published.
```

Cargo.toml
```diff
@@ -4,7 +4,7 @@ members = [
     "agent/resource-agent",
     "agent/test-agent",
     "bottlerocket/agents",
-    "bottlerocket/testsys",
+    "testsys",
```
webern: I'm open to suggestions as far as naming goes, but I think we should not use `testsys` as the name of this CLI. I don't want there to be confusion between the CLI we create in the Bottlerocket repo and the CLI in the `bottlerocket-test-system` repo. It needs to be clear that they are not the same thing!

```suggestion
    "cli",
```

webern (commented):
This is great work, I have some more suggestions.

webern:
^ I pushed a rebase

model/src/test_manager/mod.rs
ecpullen: We should discuss this. I prefer the current implementation because it aligns with `Crd`.

webern (approved):
Looks good. I changed the base branch to a new branch named`monorepo`. We will use that branch for a bit while we iterate on this.

PR #388, opened 2022-05-04
build: fix aws-iam-authenticator build

webern:
**Issue number:**

N.A

**Description of changes:**

Apparently, GitHub archive hashes are not stable. The sha sum for
aws-iam-authenticator changed and it broke our build. Now we pull it
from the Bottlerocket lookaside cache instead.

We probably need to do something like this with the rest of our GitHub archive sourcecode pulls as well, but for now this fixes our build.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -191,7 +191,8 @@ RUN mkdir -p /usr/share/licenses/aws-iam-authenticator && \
     chown -R builder:builder /usr/share/licenses/aws-iam-authenticator
 
 ARG AWS_IAM_AUTHENTICATOR_VERSION=0.5.3
-ARG AWS_IAM_AUTHENTICATOR_SOURCE_URL="https://github.com/kubernetes-sigs/aws-iam-authenticator/archive/refs/tags/v${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz"
+ARG AWS_IAM_AUTHENTICATOR_SHA512_SUM=430af9fd04b9a94205a485281fb668f5bc18cdac569de0232fa98e08ebb0e08a8d233537bd3373a5f1e53cf529bc2050aebc34a4a53c8b29a831070e34213210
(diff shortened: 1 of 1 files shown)


etungsten (approved):

Dockerfile
```diff
@@ -190,8 +190,9 @@ USER root
 RUN mkdir -p /usr/share/licenses/aws-iam-authenticator && \
     chown -R builder:builder /usr/share/licenses/aws-iam-authenticator
 
-ARG AWS_IAM_AUTHENTICATOR_VERSION=0.5.3
-ARG AWS_IAM_AUTHENTICATOR_SOURCE_URL="https://github.com/kubernetes-sigs/aws-iam-authenticator/archive/refs/tags/v${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz"
+ARG AWS_IAM_AUTHENTICATOR_VERSION=0.5.5
+ARG AWS_IAM_AUTHENTICATOR_SHA512_SUM=24c2a7fd141f921e18c17f4c7e90bbf5325cc6fad3b35753c0da2d879cf64d8f6af0130f8c45ec7693d6fce043181abbd04a911e6fa778f686bd5c313631ed0f
+ARG AWS_IAM_AUTHENTICATOR_SOURCE_URL="https://cache.bottlerocket.aws/aws-iam-authenticator-0.5.5.tar.gz/${AWS_IAM_AUTHENTICATOR_SHA512_SUM}/aws-iam-authenticator-${AWS_IAM_AUTHENTICATOR_VERSION}.tar.gz"
```
jpculp: I don't have strong feelings about this, but should we take the `sdk-fetch`/hashes approach that we're using with the other Bottlerocket containers?

jpculp (approved):

Dockerfile
webern: Hmm, probably, I would need to find a prototype example. Are these in the Bottlerocfket rpm spec files?

jpculp: This is probably the most straight forward example: https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/35

webern: I'm going to leave sdk-fetch as a future enhancement. It seems like it makes the most sense when you have all of your sources in the lookaside cache.

webern:
^ We need to use 0.5.3 to avoid a breaking change for now.

PR #420, opened 2022-05-20
test_manager: change `Status` to `StatusSnapshot` and add info

ecpullen:
**Issue number:**

N/A

**Description of changes:**

`Status` is slightly ambiguous because once it is created all data remains permanent. `StatusSnapshot` helps iterate that data may become stale.
This pr also adds 3 additional fields to the `StatusSnapshot` struct: `passed`, `finished` and `failed_tests`. These will be useful for ci when calling `testsys status --json` as the test results and determining if tests are still running will not need to be calculated.

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/cli/src/status.rs
+++ b/cli/src/status.rs
@@ -31,7 +31,7 @@ impl Status {
         } else {
             let (terminal_size::Width(width), _) =
                 terminal_size::terminal_size().unwrap_or((Width(120), Height(0)));
-            println!("{}", status.to_string(width as usize));
+            println!("{:width$}", status, width = width as usize);
         }
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


webern (approved):

cli/src/status.rs
```diff
@@ -31,7 +31,7 @@ impl Status {
         } else {
```
arnaldo2792, line 31: Nit:

Your commit subject should adhere to the following standard suggested [here](https://github.com/bottlerocket-os/bottlerocket/pull/2219#issuecomment-1176887596):

>https://cbea.ms/git-commit/
 Minor changes:
>- don't capitalize the subject
>- start with a "subsystem: blah" tag unless the change is very cross-cutting

arnaldo2792 (approved):

etungsten (approved):

PR #439, opened 2022-06-02
Add make targets for image tag and publishing

### LINKED ISSUE #436: Tag and push contianer images in Makefile target
The Makefile should have a target that allows a user to tag and push the images built with `make images`, otherwise there could be a disconnect between the images that are built and what is tagged/push to a container registry.

ecpullen:
**Issue number:**
Closes #436

**Description of changes:**

Added make targets for `tag-images` and `publish-images`. `tag-images` can be configured with `PUBLISH_IMAGES_REGISTRY`. `PUBLISH_IMAGES_VERSION` must be included regardless.

**Testing done:**
Used `tag-images` with and without a repo.
Used `publish-images` with a repo.

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,15 @@ TOOLS_IMAGE ?= public.ecr.aws/bottlerocket/bottlerocket-test-tools:$(BOTTLEROCKE
 IMAGES = controller sonobuoy-test-agent ec2-resource-agent eks-resource-agent ecs-resource-agent \
 	migration-test-agent vsphere-vm-resource-agent ecs-test-agent
 
+# Store targets for tagging images
+TAG_IMAGES = $(addprefix tag-, $(IMAGES))
+
+# Store targets to push images
+PUSH_IMAGES = $(addprefix push-, $(IMAGES))
+
 .PHONY: build sdk-openssl example-test-agent example-resource-agent \
-	images fetch integ-test show-variables cargo-deny tools $(IMAGES)
+	images fetch integ-test show-variables cargo-deny tools $(IMAGES) \
+	tag-images $(TAG_IMAGES) push-images $(PUSH_IMAGES)
 
 export DOCKER_BUILDKIT=1
 export CARGO_HOME = $(TOP)/.cargo
@@ -130,3 +137,29 @@ cargo-deny:
 	cargo install --version 0.9.1 cargo-deny --locked
 	cargo fetch
 	cargo deny --all-features --no-default-features check --disable-fetch licenses sources
+
+# Define a target to tag all images
+tag-images: $(TAG_IMAGES)
+
+# This defines the TAG_IMAGE variable, extracting the image name from the target name
+$(TAG_IMAGES): TAG_IMAGE = $(@:tag-%=%)
+$(TAG_IMAGES): check-publish-version
+	docker tag $(TAG_IMAGE) $(if $(PUBLISH_IMAGES_REGISTRY), $(PUBLISH_IMAGES_REGISTRY)/)$(TAG_IMAGE):$(PUBLISH_IMAGES_VERSION)
+
+# Define a target to tag all images
+publish-images: $(PUSH_IMAGES)
+
+# This defines the TAG_IMAGE variable, extracting the image name from the target name
+$(PUSH_IMAGES): TAG_IMAGE = $(@:push-%=%)
+$(PUSH_IMAGES): check-publish-version check-publish-registry
+	docker push $(if $(PUBLISH_IMAGES_REGISTRY), $(PUBLISH_IMAGES_REGISTRY)/)$(TAG_IMAGE):$(PUBLISH_IMAGES_VERSION)
+
+check-publish-version:
+ifndef PUBLISH_IMAGES_VERSION
+	$(error PUBLISH_IMAGES_VERSION is undefined)
+endif
+
+check-publish-registry:
+ifndef PUBLISH_IMAGES_REGISTRY
+	$(error PUBLISH_IMAGES_REGISTRY is undefined)
+endif


Makefile
```diff
@@ -16,8 +16,16 @@ TOOLS_IMAGE ?= public.ecr.aws/bottlerocket/bottlerocket-test-tools:$(BOTTLEROCKE
...
+TAG_IMAGES = $(addprefix tag-, $(IMAGES))
+
+# Store targets for tagging images
+PUSH_IMAGES = $(addprefix push-, $(IMAGES))
+
 .PHONY: build sdk-openssl example-test-agent example-resource-agent \
-	images fetch integ-test show-variables cargo-deny tools $(IMAGES)
+	images fetch integ-test show-variables cargo-deny tools $(IMAGES) \
+	tag-images $(TAG_IMAGES)
+
```
webern, line 28:
```suggestion
```

Makefile
```diff
@@ -130,3 +138,27 @@ cargo-deny:
...
 	cargo deny --all-features --no-default-features check --disable-fetch licenses sources
+
+# Define a target to tag all images
+tag-images: $(TAG_IMAGES)
+
+# This defines the TAG_IMAGE variable, extracting the image name from the target name
+$(TAG_IMAGES): TAG_IMAGE = $(@:tag-%=%)
+
+$(TAG_IMAGES): check-publish-version
+	docker tag $(TAG_IMAGE) $(if $(PUBLISH_IMAGES_REPO), $(PUBLISH_IMAGES_REPO)/)$(TAG_IMAGE):$(PUBLISH_IMAGES_VERSION)
```
webern, line 149: So the space here `$(TAG_IMAGE) $(if` is not rendered?

webern (approved):
@arnaldo2792 does this look like what you had in mind?

Makefile
```diff
@@ -130,3 +137,27 @@ cargo-deny:
...
 	cargo fetch
 	cargo deny --all-features --no-default-features check --disable-fetch licenses sources
+
+# Define a target to tag all images
+tag-images: $(TAG_IMAGES)
+
+# This defines the TAG_IMAGE variable, extracting the image name from the target name
+$(TAG_IMAGES): TAG_IMAGE = $(@:tag-%=%)
+
+$(TAG_IMAGES): check-publish-version
```
arnaldo2792: nit: Try to have these two targets together (no space in between them), same for similar targets

Makefile
```diff
@@ -16,8 +16,15 @@ TOOLS_IMAGE ?= public.ecr.aws/bottlerocket/bottlerocket-test-tools:$(BOTTLEROCKE
 IMAGES = controller sonobuoy-test-agent ec2-resource-agent eks-resource-agent ecs-resource-agent \
 	migration-test-agent vsphere-vm-resource-agent ecs-test-agent
 
+# Store targets for tagging images
+TAG_IMAGES = $(addprefix tag-, $(IMAGES))
+
+# Store targets for tagging images
```
arnaldo2792: These are the targets to push images

Makefile
```diff
@@ -16,8 +16,15 @@ TOOLS_IMAGE ?= public.ecr.aws/bottlerocket/bottlerocket-test-tools:$(BOTTLEROCKE
...
+PUSH_IMAGES = $(addprefix push-, $(IMAGES))
+
 .PHONY: build sdk-openssl example-test-agent example-resource-agent \
-	images fetch integ-test show-variables cargo-deny tools $(IMAGES)
+	images fetch integ-test show-variables cargo-deny tools $(IMAGES) \
+	tag-images $(TAG_IMAGES)
```
arnaldo2792: You are missing $(PUSH_IMAGES)

Makefile
```diff
@@ -130,3 +137,27 @@ cargo-deny:
...
+
+# Define a target to tag all images
+publish-images: $(PUSH_IMAGES)
+
+# This defines the TAG_IMAGE variable, extracting the image name from the target name
+$(PUSH_IMAGES): TAG_IMAGE = $(@:push-%=%)
+
+
+$(PUSH_IMAGES): check-publish-version
+	docker push $(if $(PUBLISH_IMAGES_REPO), $(PUBLISH_IMAGES_REPO)/)$(TAG_IMAGE):$(PUBLISH_IMAGES_VERSION)
```
arnaldo2792: For pushing images, I think it makes sense to make `PUBLISH_IMAGES_REPO` required. With the current approach, the images will be pushed to the docker hub (provided that the user has logged in) since that is the default registry for `docker push`.

webern: > For pushing images, I think it makes sense to make PUBLISH_IMAGES_REPO required. With the current approach, the images will be pushed to the docker hub (provided that the user has logged in) since that is the default registry for docker push.

Yeah that's my fault, I defined the requirements and figured it would fail if you accidentally tried to push to dockerhub unless you had created all of those repos there. But I think you're right, it would be better to require the registry explicitly.

Also @ecpullen I think we should change it to `PUBLISH_IMAGES_REGISTRY`. I had the terminology wrong when we chatted about this. the URI parts are `registry/repo:version`

arnaldo2792:
I think you may have added another commit unrelated to the original PR? I might be wrong

ecpullen:
Yes I did that so we can merge both prs without me having to rebase them.

webern:
> Yes I did that so we can merge both prs without me having to rebase them.

Unfortunately I was going to merge this but can't because of this!

arnaldo2792 (approved):

webern:
> Unfortunately I was going to merge this but can't because of this!

I was able to rebase and drop the extra commit so this will be merge-able after CI.

PR #447, opened 2022-06-14
Allow eksctl config in eks resource provider

### LINKED ISSUE #252: eks: add options for creating and testing Bottlerocket in EKS IPv6 clusters
We need to be able to validate Bottlerocket's IPv6 capabilities in an IPv6 EKS cluster.

The v0.79.0 of eksctl supports IPv6 cluster creation, we should add that functionality to the eks-resource-agent.
The ec2-resource-agent would also need to then know the ipFamily of the cluster the node needs to connect to and write out its userdata `kubernetes.settings` accordingly

See https://eksctl.io/usage/vpc-ip-family/

### LINKED ISSUE #416: eks provider: set up aws config map
We need to be able to add desired roles and users to the AwsConfig map when we create an EKS cluster. For example, if the testsys agents themselves are running on EKS, then the target EKS clusters need to allow control from the testsys cluster node's role.

This might be related to #225

ecpullen:
**Issue number:**

Closes #252 
Closes #416 

**Description of changes:**

Allows the user to provide and encoded eksctl config file for cluster creation.
I chose to keep the config flat (instead of creating an enum for cluster config vs zones and version) so that the work is compatible with #537 .

**Testing done:**

Tested using existing roles from another cluster with the following config: (Note: this config is probably not minimal but it works)
```yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig
metadata:
  name:  eksctl-test
  region: us-west-2
  version: "1.22"
kubernetesNetworkConfig:
  ipFamily: IPv4
iam:
  serviceRoleARN: arn:aws:iam::<ACCOUNT-ID>:role/eksctl-x86-64-aws-k8s-121-cluster-ServiceRole-...
iamIdentityMappings:
  - arn: arn:aws:iam::<ACCOUNT-ID>:role/Ecs
    groups:
      - system:masters
    username: admin
    noDuplicateARNs: true
nodeGroups:
  - name: empty
    desiredCapacity: 0
    iam:
      withAddonPolicies:
        albIngress: false
        appMesh: false
        appMeshPreview: false
        autoScaler: false
        awsLoadBalancerController: false
        certManager: false
        cloudWatch: false
        ebs: false
        efs: false
        externalDNS: false
        fsx: false
        imageBuilder: false
        xRay: false
      instanceRoleARN: "arn:aws:iam::<ACCOUNT-ID>:role/eksctl-x86-64-aws-k8s-121-nodegro-NodeInstanceRole-..."
      instanceProfileARN: "arn:aws:iam::<ACCOUNT-ID>:instance-profile/eksctl-x86-64-aws-k8s-121-nodegroup-ng-..."
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -7,7 +7,7 @@ use aws_sdk_eks::model::{Cluster, IpFamily};
 use aws_sdk_eks::output::DescribeClusterOutput;
 use aws_types::SdkConfig;
 use bottlerocket_types::agent_config::{
-    CreationPolicy, EksClusterConfig, K8sVersion, AWS_CREDENTIALS_SECRET_NAME,
+    CreationPolicy, EksClusterConfig, EksctlConfig, K8sVersion, AWS_CREDENTIALS_SECRET_NAME,
 };
 use log::{debug, info, trace};
 use model::{Configuration, SecretName};
@@ -18,6 +18,7 @@ use resource_agent::provider::{
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::env::temp_dir;
+use std::fs::write;
 use std::path::Path;
 use std::process::Command;
 
@@ -25,6 +26,7 @@ use std::process::Command;
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
 const DEFAULT_VERSION: &str = "1.21";
+const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
 
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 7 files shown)


bottlerocket/types/src/agent_config.rs
```diff
@@ -104,6 +104,9 @@ pub struct EksClusterConfig {
     /// quoted so that it is interpreted as a JSON/YAML string (not a number).
     pub version: Option<K8sVersion>,
 
+    /// The encoded eksctl config that should be used for cluster creation.
+    pub encoded_eksctl_config: Option<String>,
```
webern: There's a problem here because the encoded_eksctl_config is exclusive to all the other configs we have in the struct. i.e. if I specify `cluster_name` and `encoded_eksctl_config` at the same time, then the `cluster_name` field has no effect, right? I wonder if we should make the EksClusterConfig an enum so that it can only be one way or the other:

```rust
enum EksClusterConfig {
    Structured(EksctlArguments),
    EncodedEksctlConfig(String),
}

Comment (ecpullen) [reply bottlerocket/types/src/agent_config.rs]: We still need the cluster name if we use eksctl config for looking up cluster information.

**Comment by ecpullen** [comment]
This pr is now ready to go.

File: bottlerocket/testsys/src/run_aws_k8s.rs [hunk 2]
```diff
@@ -102,6 +102,10 @@ pub(crate) struct RunAwsK8s {
     #[structopt(long)]
     cluster_provider_pull_secret: Option<String>,
 
+    /// Path to eksctl config file.
+    #[structopt(long)]
+    cluster_config_path: Option<String>,
```
etungsten: To keep it consistent with `sonobuoy_e2e_repo_config`
```suggestion
    cluster_config: Option<String>,
```

bottlerocket/types/src/agent_config.rs
```diff
@@ -92,12 +92,15 @@ impl Configuration for MigrationConfig {}
 #[serde(rename_all = "camelCase")]
 pub struct EksClusterConfig {
     /// The name of the eks cluster to create or an existing cluster.
+    /// If an eksctl config is passed, the names must match.
```
etungsten: Below, the docstring for `encoded_eksctl_config` says 

>This conflicts with `cluster_name`, `zones` and `version`. If an encoded cluster config is provided the other values will not be used.

So this wouldn't be accurate right? Can we turn `cluster_name` into an option? I'd rather have `encoded_eksctl_config` be the source of truth if it's specified.

bottlerocket/types/src/agent_config.rs
etungsten: Would it be possible to derive the cluster name from the cluster config if it's specified? Extract it out with `serde_yaml::Value` maybe? We already pull in `serde_yaml` as a dependency elsewhere in testsys.

bottlerocket/types/src/agent_config.rs
```diff
@@ -92,12 +92,15 @@ impl Configuration for MigrationConfig {}
...
     pub cluster_name: String,
 
     /// Whether this agent will create the cluster or not.
     pub creation_policy: Option<CreationPolicy>,
 
-    /// The AWS region to create the cluster. If no value is provided `us-west-2` will be used.
+    /// The AWS region to create the cluster. If no value is provided `us-west-2` will be used. If
+    /// an eksctl cluster is provided, this region should match the one specified as this is the
```
etungsten: This is slightly vague. Is "an eksctl cluster" referring to an existing EKS cluster, or an eksctl cluster config (the same messaging as the comment above for `cluster_name`?)?

This should be treated the same as `cluster_name` or anything that can specified in the eksctl cluster config since that should always have `region` included in there.

bottlerocket/types/src/agent_config.rs
ecpullen: The only way to do that would be to serialize the encoded config. Since we use eksctl and the aws sdk we have to know the cluster name.

bottlerocket/types/src/agent_config.rs
etungsten: I think to take the idea even further, we should consider making the eksctl cluster configuration file mandatory if cluster creation policy is `always` or `ifnotexists`, and the set of other fields are only used if cluster creation policy is never. 

We might need to turn `EksClusterConfig` itself into an enum, either we represent an existing cluster with `cluster_name`, `region`. Or we represent an EKS cluster that has yet to be created with an eksctl cluster config. I don't believe we really need the "create if not exist" flow (is there?). 

All of this non-blocking, but I do think it would clear things up a lot better.

ecpullen: Requiring an eksctl config just makes it harder to start using the framework. Not only does a user need to learn TestSys but now they also need to learn eksctl configs.

bottlerocket/types/src/agent_config.rs
etungsten: I don't think it'd be that much overhead to deserialize the encoded cluster config and extract out the `cluster_name`, `region`. If they don't exist (they should), then we fallback to defaults or error. Like I mentioned in another comment, we're already using `serde_yaml` elsewhere.

bottlerocket/types/src/agent_config.rs
etungsten: After thinking about it further, I think the "create if not exist" workflow is useful and we need the current interface to make that possible. 

I still strongly feel that we shouldn't have any field that, when specified, must match another value elsewhere. In this case, cluster name/region and the corresponding fields in `eksctl`'s cluster config. As much as I dislike implicit overrides in configuration, there really needs be a single source of truth that takes precedence over other fields. Take `eksctl create cluster` for example, it doesn't let you specify any flag if a configuration file is specified. Ideally we would match that behavior but it's harder to enforce through the TestSys agent configuration.

ecpullen: I have updated the interface to use the eksctl config as the source of truth if it is provided. Otherwise, it will use provided fields. Now there is no duplication required. Running a few tests and then I'll push.

ecpullen:
^ Use eksctl config to determine the cluster name and region instead of provided arguments.
Verified with `testsys run aws-k8s`
```
 NAME                    TYPE        STATE        PASSED    SKIPPED    FAILED
 eks-test                Test        passed       1         6441       0
 my-cluster              Resource    completed
 my-cluster-instances    Resource    completed
 ```

**Review by stmcginnis — APPROVED** [review]
This looks fine to me, but would be good to have someone with more experience in this area take a look.

File: bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs [hunk 5]
```diff
@@ -108,6 +110,180 @@ impl AwsClients {
...
+                            "The cluster's region was not in the eksctl config.",
+                        )?
+                        .to_string(),
+                }
+            }
+            (Some(cluster_name), None) => Self::Args {
+                cluster_name: cluster_name.to_string(),
+                region: region
+                    .as_ref()
+                    .cloned()
```
etungsten: Can this be `to_owned()`?
```suggestion
                    .to_owned()
```

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -177,20 +181,21 @@ impl RunAwsK8s {
         let cluster_resource_name = self
             .cluster_resource_name
             .as_ref()
-            .unwrap_or(&self.cluster_name);
+            .cloned()
```
etungsten, line 184: Similar to above, is the `as_ref()` still necessary?

etungsten (approved):

ecpullen:
^ Change `.as_ref().clone()` to `to_owned()`

etungsten (approved):

ecpullen:
^ Clippy fixes

webern (changes requested):
I'm trying to remember where we left off with this. I think I had asked you to revisit the possibility of using an enum so that the ambiguity of args-vs-configfile would be impossible. I believe you said that it had proved difficult before but that you would look at it again.

We don't want this PR to block forever, but the interface is pretty bad as is so it's probably worth another look to see if we can make the input "always correct" by design.

ecpullen:
^ Uses an enum instead of a flat config.

ecpullen:
^ Rebase onto develop

ecpullen:
^ Clippy fixes.

bottlerocket/types/src/agent_config.rs
```diff
@@ -91,24 +91,58 @@ impl Configuration for MigrationConfig {}
...
-    /// The eks version of the the cluster (e.g. "1.14", "1.15", "1.16"). Make sure this is
-    /// quoted so that it is interpreted as a JSON/YAML string (not a number).
-    pub version: Option<K8sVersion>,
+    // /// The eks version of the the cluster (e.g. "1.14", "1.15", "1.16"). Make sure this is
+    // /// quoted so that it is interpreted as a JSON/YAML string (not a number).
+    // pub version: Option<K8sVersion>,
 
+    // /// The encoded eksctl config that should be used for cluster creation. This conflicts with
+    // /// `cluster_name`, `zones` and `version`. If an encoded cluster config is provided the other
+    // /// values will not be used.
```
webern: clean up commented code before merging

bottlerocket/types/Cargo.toml
```diff
@@ -9,3 +9,4 @@ license = "MIT OR Apache-2.0"
 model = { version = "0.0.2", path = "../../model" }
 serde = "1"
 serde_plain = "1"
+serde_yaml = "0.8"
```
webern:
```suggestion
serde_yaml = "0.8"

```

webern (approved):
Nice! Thank you for the extra effort to get the enum to work. I think it makes it much more clear.

ecpullen:
^ Code cleanup

PR #450, opened 2022-06-15
Adds uninstall capabilities to the TestManager

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Allows users to uninstall the k8s components created by `install`.

**Testing done:**

Used `cli uninstall` to uninstall from a cluster and  everything was properly cleaned up.

**Terms of contribution:**

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -14,6 +14,7 @@ mod results;
 mod run;
 mod run_file;
 mod status;
+mod uninstall;
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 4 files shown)


webern (approved):
I think the CLI should "delete --all" before running this because it will be less messy if something fails its delete phase. But the TestManager doesn't need to do it for the CLI so this is good.

cbgbt (approved):

zmrow (approved):
:microphone:

PR #470, opened 2022-07-11
agent: fix task state during test retries.

ecpullen:
**Issue number:**
N/A

**Description of changes:**

Fixes a bug where the task state of a test was set as `Completed` even if the test was being rerun. Previously the `TaskState` was set to complete as soon as a the first test result was in even though the test may be rerunning. This pr stops the `TaskState` from changing until the test is complete.

**Testing done:**

Ran a test that failed and was rerun and `task_state` was not set to complete.

**Terms of contribution:**

### DIFF
--- a/agent/test-agent/src/agent.rs
+++ b/agent/test-agent/src/agent.rs
@@ -106,7 +106,7 @@ where
             );
             if let Err(e) = self
                 .client
-                .send_test_done(test_results.clone())
+                .send_test_results(test_results.clone())
                 .await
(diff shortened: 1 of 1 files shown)


etungsten (approved):

arnaldo2792:
Just for my own sanity, could you please update the description and describe how did you test your change and validated the bug was fixed?

stmcginnis (approved):
Looks good - line 134 will mark the test done at the correct time, so makes sense to only update the results here.

PR #476, opened 2022-07-18
controller: Allow exclusive access to resources

### LINKED ISSUE #428: controller: support updown on same cluster
Currently upgrade-downgrade testing (in which we launch a Bottlerocket instance, do an upgrade into a new version, then downgrade back to the original version) requires access to an EKS cluster that is exclusive. Sonobuoy conformance testing also requires exclusive access (and these exclusivity requirements may also apply to ECS as well).

Here's why:
Sonobuoy testing uses all the nodes on a cluster. So we need to make sure all of the nodes on the cluster are running the Bottlerocket OS that we want to test.
We can run conformance testing with a few nodes or we can run upgrade downgrade (*updown*) testing, but not both at the same time.

Let's say we are testing bottlerocket version 1.17.1 and the previous version is 1.17.0...

We want a way to model this (CRDs) and have the controller do the following:
- create cluster (if desired)
- run nodes for conformance testing of v1.17.1
- run conformance test
- terminate the nodes, but not the cluster
- run nodes of version 1.17.0
- run the updown test that will upgrade these to 1.17.1 and back to 1.17.0 and check them with sonobuoy quick

Then we will delete the test and resources.

We should be able to create all of these resource and test CRD objects at once, expressing the above desired execution path, and have the controller do it.

Note there may be additional exclusivity requirements for different/new testing modalities (see `#TODO`), but updown is a known/scoped case that we have before us now even before we write new test modalities such as #419 and #426.

### LINKED ISSUE #429: controller: support exclusive access to bottlerocket nodes
Similar to #428, when we start introducing certain special Bottlerocket tests, these tests may require exclusive access to a Bottlerocket host, or they may need to run on Bottlerocket hosts that have been reset or otherwise are in a state appropriate for a given test. This issue represents the expectation that there will be requirements expending what we have done in #428

Considered an extension of functionality that is more specialized in #428 and thus...

depends on
- #428

ecpullen:
**Issue number:**
Closes #429 
Closes #428

**Description of changes:**

Adds a new field `conflicts_with` to the `resource` crd. This field allows users to prevent a resource from being created while another resource still exists. 

In order to prevent the user from being required to manually delete the resource, 2 new `DestructionPolicy`s were implemented: `OnTestSuccess` and `OnTestCompletion`. `OnTestSuccess` will delete the resource once all resources depending on it have been deleted and all tests requiring it passed. `OnTestCompletion` will delete the resource once all resources depending on it are deleted and all tests using it have completed regardless of test outcome. 

This significantly reduces the waiting time when testing bottlerocket because instances will be cleaned up as soon as possible and not at the command of the operator.

**Testing done:**

Tested multiple resources requiring the same resource and the resource with `conflicts_with` did not start until the other resources had been deleted.

Tested that resources waiting for depending resources to be deleted before starting their own automatic deletion.

Tested that resources properly delete based on `OnTestSuccess` and `OnTestCompletion`.

**Terms of contribution:**

### DIFF
--- a/controller/src/resource_controller/action.rs
+++ b/controller/src/resource_controller/action.rs
@@ -1,11 +1,12 @@
 use crate::error::Result;
 use crate::job::{JobState, TEST_START_TIME_LIMIT};
 use crate::resource_controller::context::ResourceInterface;
+use kube::core::object::HasSpec;
 use kube::ResourceExt;
 use log::{debug, trace};
-use model::clients::CrdClient;
+use model::clients::{AllowNotFound, CrdClient, TestClient};
 use model::constants::{FINALIZER_CREATION_JOB, FINALIZER_MAIN, FINALIZER_RESOURCE};
-use model::{CrdExt, DestructionPolicy, ResourceAction, TaskState};
+use model::{CrdExt, DestructionPolicy, ResourceAction, TaskState, TestUserState};
 use parse_duration::parse;
 
 /// The action that the controller needs to take in order to reconcile the [`Resource`].
@@ -22,6 +23,7 @@ pub(super) enum CreationAction {
     AddJobFinalizer,
     StartJob,
     WaitForDependency(String),
+    WaitForConflict(String),
     WaitForCreation,
     AddResourceFinalizer,
     Done,
@@ -30,6 +32,7 @@ pub(super) enum CreationAction {
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub(super) enum DestructionAction {
+    StartResourceDeletion,
     RemoveCreationJob,
     RemoveCreationJobFinalizer,
     StartDestructionJob,
@@ -52,7 +55,7 @@ pub(super) enum ErrorState {
 }
 
 pub(super) async fn action(r: &ResourceInterface) -> Result<Action> {
-    if r.resource().is_delete_requested() {
+    if r.resource().is_delete_requested() || is_deletion_required(r).await? {
         Ok(Action::Destruction(destruction_action(r).await?))
     } else {
         Ok(Action::Creation(creation_action(r).await?))
@@ -69,6 +72,10 @@ async fn creation_action(r: &ResourceInterface) -> Result<CreationAction> {
     }
 
     if let Some(wait_action) = dependency_wait_action(r).await? {
+        return Ok(wait_action);
+    }
+
+    if let Some(wait_action) = conflicting_wait_action(r).await? {
         return Ok(wait_action);
     }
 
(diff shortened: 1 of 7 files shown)


controller/src/resource_controller/action.rs
```diff
@@ -151,8 +182,52 @@ async fn creation_completed_action(r: &ResourceInterface) -> Result<CreationActi
     }
 }
 
+/// Destruction for a resource may be required if the resources `DestructionPolicy` is
+/// `OnTestSuccess` or `OnTestCompletion`.
+async fn deletion_needed(r: &ResourceInterface) -> Result<bool> {
```
etungsten: nit: we already have `is_deletion_requested`, so maybe
```suggestion
async fn is_deletion_required(r: &ResourceInterface) -> Result<bool> {
```

controller/src/resource_controller/action.rs
```diff
@@ -151,8 +182,52 @@ async fn creation_completed_action(r: &ResourceInterface) -> Result<CreationActi
...
+                        }
+                    }
+                    _ => (),
+                }
+            }
+        }
+        Ok(true)
+    } else {
+        Ok(false)
+    }
```
etungsten: nit: 
```suggestion
    }
    Ok(false)
```

controller/src/resource_controller/action.rs
```diff
@@ -151,8 +182,52 @@ async fn creation_completed_action(r: &ResourceInterface) -> Result<CreationActi
...
 }
 
+/// Destruction for a resource may be required if the resources `DestructionPolicy` is
+/// `OnTestSuccess` or `OnTestCompletion`.
+async fn deletion_needed(r: &ResourceInterface) -> Result<bool> {
+    let destruction_policy = r.resource().spec.destruction_policy;
+    if matches!(
+        destruction_policy,
+        DestructionPolicy::OnTestCompletion | DestructionPolicy::OnTestSuccess
+    ) {
```
etungsten: I prefer a check to early return here so we can un-indent most of this logic so it's not as nested. The check on resource dependencies after this is also independent of the destruction policy.

etungsten (commented):
For future reference, I think it's better in this case to separate these changes into two commits, one for adding `conflicts_with` and the other for adding the new destruction policies.

etungsten (approved):

controller/src/resource_controller/action.rs
```diff
@@ -151,8 +182,51 @@ async fn creation_completed_action(r: &ResourceInterface) -> Result<CreationActi
...
+        return Ok(false);
+    }
+    // If any resources still depend on this resource it should not be deleted.
+    let resources = r.resource_client().get_all().await?;
+    for resource in resources {
+        if let Some(depends_on) = resource.spec().depends_on.as_ref() {
+            if depends_on.contains(&r.name().to_string()) {
+                return Ok(false);
```
stmcginnis, line 200: Still just learning the code, so may be a dumb question - if this resource is marked `OnTestCompletion` or `OnTestSuccess`, but it is a dependency for another resource yet, how/where does it eventually get deleted once that dependent has been destroyed?

ecpullen: Currently it doesn't until it is deleted itself. With this pr if the `DestructionPolicy` is `OnTestCompletion` or `OnTestSuccess` then it will be deleted once the resources depending on it are deleted.

stmcginnis (approved):

PR #500, opened 2022-08-12
Add help target to Makefile

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a `help` target and sets it as the default. This helps newer
contributors discover the make targets of relevance by providing a
nicely formatted output of the targets they might be interested in.

This can be expanded as we go to add additional targets or to categorize
targets into different groups based on their expected usage.

Also some minor cleanup to the Makefile itself.

**Testing done:**

image

**Terms of contribution:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,10 @@
+# Set our default to print out help if user has not provided a target
+.DEFAULT_GOAL := help
+
+# Determine our root directory path for later use
 TOP := $(dir $(firstword $(MAKEFILE_LIST)))
 
-TESTSYS_BUILD_HOST_UNAME_ARCH=$(shell uname -m)
-TESTSYS_BUILD_HOST_GOARCH ?= $(lastword $(subst :, ,$(filter $(TESTSYS_BUILD_HOST_UNAME_ARCH):%,x86_64:amd64 aarch64:arm64)))
-TESTSYS_BUILD_HOST_PLATFORM=$(shell uname | tr '[:upper:]' '[:lower:]')
-# On some hosts we get an x509 certificate error and need to set GOPROXY to "direct"
-TESTSYS_BUILD_GOPROXY ?= direct
-
+# Variables we update as newer versions are released
 BOTTLEROCKET_SDK_VERSION = v0.26.0
 BOTTLEROCKET_SDK_ARCH = $(TESTSYS_BUILD_HOST_UNAME_ARCH)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,18 +1,26 @@
...
 TOOLS_IMAGE ?= public.ecr.aws/bottlerocket-test-system/bottlerocket-test-tools:$(BOTTLEROCKET_TOOLS_VERSION)
 
+# Capture information about the build host
+TESTSYS_BUILD_HOST_UNAME_ARCH=$(shell uname -m)
+TESTSYS_BUILD_HOST_GOARCH ?= $(lastword $(subst :, ,$(filter $(TESTSYS_BUILD_HOST_UNAME_ARCH):%,x86_64:amd64 aarch64:arm64)))
+TESTSYS_BUILD_HOST_PLATFORM=$(shell uname | tr '[:upper:]' '[:lower:]')
+# On some hosts we get an x509 certificate error and need to set GOPROXY to "direct"
+TESTSYS_BUILD_GOPROXY ?= direct
+
+# The set of images to create. Add new artifacts here when added to the
```
ecpullen: Maybe note that these are for bottlerocket images and the example images have their own targets.

Makefile
```diff
@@ -153,15 +165,15 @@ cargo-deny:
 	cargo deny --all-features --no-default-features check --disable-fetch licenses sources
 
 # Define a target to tag all images
-tag-images: $(TAG_IMAGES)
+tag-images: $(TAG_IMAGES)  ## tag all images
```
ecpullen, line 168: Should we put the env variables for registry and tag?

ecpullen (approved):
LGTM

Makefile
stmcginnis: Not sure I understand, can you elaborate?

ecpullen: The registry can be configured for tagging with `PUBLISH_IMAGES_REGISTRY` and the tag can be configured with `PUBLISH_IMAGES_VERSION`

stmcginnis: That gets to be a bit long to add detailed help text. Do you mind if we start with the more concise help output and look at better ways of providing these details as a follow on?

stmcginnis:
Rebased on `develop` to pick up the latest updates.

etungsten (approved):

PR #502, opened 2022-08-12
Truncate resource name and append object UID

### LINKED ISSUE #353: Resource name length limit
If a resources name is too long, a job will not be created for destruction even though a job could be created for resource creation. (`creation` has fewer characters than `destruction`).

mjsterckx:
**Issue number:**

Closes #353 

**Description of changes:**

Instead of simply appending -creation and -destruction to the resource name to generate the job names, the object UID created by k8s is now appended to a truncated resource name. This ensures that every creation- and destruction-job has a unique name, even if the resource names contain the same first 51 or more characters. Since the resource name gets truncated down to 15 characters, the situation where a destruction pod cannot be created because the name is too long will never occur.

The does_resource_destruction_pod_exist test in cluster.rs has been changed to reflect the new destruction pod naming scheme.

**Testing done:**

- Ran testsys with a test YAML file that depends on one resource with a name shorter than 51 characters
- Ran testsys with a test YAML file that depends on one resource with a name between 51 characters and 54 characters
- Ran testsys with a test YAML file that depends on one resource with a name longer than 54 characters
- Ran testsys with a test YAML file that depends on two resources with names longer than 51 characters, where the first 51 characters are the same
- Checked that all of the above resources could be created and deleted successfully, and the test passed

**Terms of contribution:**

### DIFF
--- a/controller/src/resource_controller/context.rs
+++ b/controller/src/resource_controller/context.rs
@@ -41,8 +41,16 @@ pub(super) struct ResourceInterface {
 
 impl ResourceInterface {
     pub(super) fn new(resource: Resource, context: Context) -> Result<Self> {
-        let creation_job = format!("{}-creation", resource.object_name());
-        let destruction_job = format!("{}-destruction", resource.object_name());
+        let mut resource_name = resource.object_name().to_string();
+
+        // Selects the UID from the k8s object to append to the truncated job name, preventing duplicate job names
+        let id = match resource.metadata.uid.as_ref() {
+            Some(r) => r,
+            None => "",
+        };
+        resource_name.truncate(15);
+        let creation_job = format!("{}{}-creation", resource_name, id);
+        let destruction_job = format!("{}{}-destruction", resource_name, id);
         Ok(Self {
             resource,
             context,
--- a/controller/src/resource_controller/context.rs
+++ b/controller/src/resource_controller/context.rs
@@ -4,7 +4,7 @@ use anyhow::Context as AnyhowContext;
 use kube::Api;
 use log::debug;
 use model::clients::{CrdClient, ResourceClient};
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):
Ran through a test run with these changes and everything looks fine. Changes look good to me.

controller/src/resource_controller/context.rs
```diff
@@ -41,8 +41,16 @@ pub(super) struct ResourceInterface {
...
-        let creation_job = format!("{}-creation", resource.object_name());
-        let destruction_job = format!("{}-destruction", resource.object_name());
+        let mut resource_name = resource.object_name().to_string();
+
+        // Selects the UID from the k8s object to append to the truncated job name, preventing duplicate job names
+        let id = match resource.metadata.uid.as_ref() {
+            Some(r) => r,
+            None => "",
+        };
+        resource_name.truncate(15);
```
jpmcb: Should this be some top level constant instead of an inline literal?

```rust
const TRUNC_LEN: i32 = 15;
```
and then used here as
```rust
resource_name.truncate(TRUNC_LEN);
```

jpmcb: I wonder if this line deserves it's own comment 🤔 

if i was reading this for the first time, I'd wonder why we needed to truncate the name of the resource down to 15 chars.

mjsterckx: Good call. I thought about it when modifying the test but decided to leave it as it was. I realize that using a constant is best practice though, so I addressed that in the new commit.

mjsterckx: Added a comment to explain why the truncating happens.

ecpullen (approved):

etungsten (approved):

PR #509, opened 2022-08-16
Removes node affinity for x86. Enables arm building/running.

### LINKED ISSUE #90: build: support arm
Support arm everywhere. The system should work on ARM and mixed x86_64/ARM TestSys clusters. Tracking issue.

jpmcb:
### Issue number:

Fixes #90 

### Description of changes:

This removes the node affinity for x86 kubernetes nodes. All images are currently buildable on 

Validated images are buildable and can be run on aarch64 dev box.

### Testing done:

Prior to this change, tests would not be scheduled on local kind clusters on an aarch64 dev-box.

Now, all images are buildable and `make integ-test` passes. I was a bit surprised this was the only change that needed to be made since most of the image infrastructure / docker commands were already in place, so if there are more validations to be done, let me know!!

**Terms of contribution:**

### DIFF
--- a/model/src/system/controller.rs
+++ b/model/src/system/controller.rs
@@ -195,20 +195,11 @@ pub fn controller_deployment(
                             required_during_scheduling_ignored_during_execution: Some(
                                 NodeSelector {
                                     node_selector_terms: vec![NodeSelectorTerm {
-                                        match_expressions: Some(vec![
-                                            NodeSelectorRequirement {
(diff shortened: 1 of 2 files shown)


jpmcb:
Force pushed to address `cargo fmt`

ecpullen (approved):
LGTM!

stmcginnis (approved):
Looks reasonable to me.

etungsten (approved):

PR #512, opened 2022-08-16
Updated the bin used for the integ-test to cli instead of testsys

### LINKED ISSUE #508: test: Convert integ test to `cli` instead of `testsys`
We will be removing the `testsys` cli from the project in the near future, so testing should use `cli` instead.

mjsterckx:
**Issue number:**

Closes #508 

**Description of changes:**

Changed the bin that is called in the integ-test from `testsys` to `cli` because `testsys` will be removed in a future update. `cli` is a more updated version.

**Testing done:**

Ran the `make integ-test` command before the changes and after, capturing all output in two files. Looked at the diff between the files and the only differences were runtimes and hashes, so the tests still perform exactly the same.

**Terms of contribution:**

### DIFF
--- a/cli/tests/cli_test.rs
+++ b/cli/tests/cli_test.rs
@@ -11,9 +11,9 @@ use tokio::time::Duration;
 const POD_TIMEOUT: Duration = Duration::from_secs(300);
 
 /// We will test:
-/// `testsys install`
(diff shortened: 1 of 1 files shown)


ecpullen (approved):

stmcginnis (approved):
Looks good!

PR #513, opened 2022-08-16
resource_controller: fix pod search filter

### LINKED ISSUE #511: Resource logs cannot be retrieved using cli
After the changes in https://github.com/bottlerocket-os/bottlerocket-test-system/commit/28142ff254c6fb081708aaf97dd70bfe13b7ea98 , the `cli logs --resource <resource-name> --state creation` (or `--state destruction`) no longer works. It simply shows the output "Unable to get logs." This happens because the TestManager still uses the old naming scheme to retrieve the resource pods (<resource-name>-creation and <resource-name>-destruction), and since those pods don't exist, it cannot find any logs.

mjsterckx:
Fixed the resource_pod method to search for the pod based on the new naming scheme (truncated name + uid + suffix) instead of name + suffix

**Issue number:**

Closes #511 

**Description of changes:**

The logs for resources could not be retrieved via the `cli logs --resource <resource-name> --state <resource-state>` command because the TestManager looked for the resource pod using the naming scheme `<resource-name>-<resource-state>`. After the changes in https://github.com/bottlerocket-os/bottlerocket-test-system/commit/28142ff254c6fb081708aaf97dd70bfe13b7ea98 , it should be looking for the resource pod based on the naming scheme `<truncated-resource-name><resource-uid>-<resource-state>`. These changes in the code allow the TestManager to find the correct pod by using the resource name to retrieve the CRD, reading the UID from there, and using that to generate the pod name. With the correct pod, the logs can be sent to the CLI.

Consolidated truncation logic into `job_name` method in `resource.rs` to avoid unnecessary API calls.

**Testing done:**

- Ran the `cli logs --resource <resource-name> --state <resource-state>` command with two names of existing resources and both `creation` and `destruction` states. In all of these cases, the logs were successfully retrieved
- Ran the `cli logs --resource <resource-name> --state <resource-state>` command with the name of a resource that did not exist. The CLI correctly displays "Unable to get logs."
- Ran all tests from https://github.com/bottlerocket-os/bottlerocket-test-system/pull/502 to make sure that objects with long/similar names can still be created and destroyed without problems

**Terms of contribution:**

### DIFF
--- a/model/src/test_manager/manager_impl.rs
+++ b/model/src/test_manager/manager_impl.rs
@@ -1,6 +1,6 @@
 use super::{error, ResourceState, Result, TestManager};
 use crate::clients::{AllowNotFound, CrdClient};
-use crate::constants::{LABEL_COMPONENT, NAMESPACE};
+use crate::constants::{LABEL_COMPONENT, NAMESPACE, TRUNC_LEN};
 use crate::{Crd, CrdName, Resource, Test};
 use k8s_openapi::api::core::v1::Pod;
 use kube::api::{ListParams, Patch, PatchParams, PostParams};
@@ -214,19 +214,42 @@ impl TestManager {
             ResourceState::Creation => "creation",
             ResourceState::Destruction => "destruction",
         };
-        pod_api
-            .list(&ListParams {
-                label_selector: Some(format!("job-name={}-{}", resource.into(), suffix)),
-                ..Default::default()
-            })
-            .await
-            .context(error::KubeSnafu { action: "get pods" })?
-            .items
-            .first()
-            .context(error::NotFoundSnafu {
-                what: "pod for test",
-            })
-            .map(|pod| pod.clone())
+        let mut resource_name = resource.into();
+        let resource_api = Api::<Resource>::namespaced(self.k8s_client.clone(), NAMESPACE);
+        let resource_crd = resource_api.get(&resource_name).await;
+        let result = match resource_crd {
+            // if the resource exists, retrieve the pod based on the truncated resource name + UID + resource state
+            Ok(resource_object) => {
+                resource_name.truncate(TRUNC_LEN);
+                return pod_api
+                    .list(&ListParams {
+                        label_selector: Some(format!(
+                            "job-name={}{}-{}",
+                            resource_name,
+                            resource_object
+                                .metadata
+                                .uid
+                                .as_ref()
+                                .unwrap_or(&"".to_string()),
+                            suffix
+                        )),
+                        ..Default::default()
+                    })
+                    .await
+                    .context(error::KubeSnafu { action: "get pods" })?
+                    .items
+                    .first()
+                    .context(error::NotFoundSnafu {
+                        what: "pod for test",
+                    })
+                    .map(|pod| pod.clone());
+            }
+            // if the resource does not exist, return an error
+            Err(_) => Err(error::Error::NotFound {
+                what: "pod for test".to_string(),
+            }),
+        };
+        result
     }
 
     /// Get a pod for the testsys controller.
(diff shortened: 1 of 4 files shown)


model/src/test_manager/manager_impl.rs
```diff
@@ -214,19 +214,42 @@ impl TestManager {
...
+                            "job-name={}{}-{}",
+                            resource_name,
+                            resource_object
+                                .metadata
+                                .uid
+                                .as_ref()
+                                .unwrap_or(&"".to_string()),
+                            suffix
+                        )),
+                        ..Default::default()
```
ecpullen: Let's convert this to a job name function (resource struct or resource client). (Also change the truncate in previous pr).

model/src/test_manager/manager_impl.rs
```diff
@@ -214,19 +214,42 @@ impl TestManager {
...
-            .await
-            .context(error::KubeSnafu { action: "get pods" })?
-            .items
-            .first()
-            .context(error::NotFoundSnafu {
-                what: "pod for test",
-            })
-            .map(|pod| pod.clone())
+        let mut resource_name = resource.into();
+        let resource_api = Api::<Resource>::namespaced(self.k8s_client.clone(), NAMESPACE);
```
ecpullen: `ResourceClient` can be used for this.

model/src/test_manager/manager_impl.rs
mjsterckx: When you say to change the truncate in previous pr, does that only refer to the truncate in `does_resource_destruction_pod_exist` from `cluster.rs`? The other one is in `new` in `context.rs` but to me it doesn't make sense to remove it there because the new `get_job_name` function will retrieve the resource CRD from the API and `new` already has the resource CRD.

ecpullen: It would be nice to have all of the truncate logic in a single function that is called whenever the resource's job/pod name is needed. Probably belongs in `resource.rs`. Then we can call `resource.job_name()` whenever we want the job name (without api calls).

mjsterckx: Unless I'm misunderstanding something, I think the places where the job names need to be generated don't have access to the Resource struct. For example, when the command `cli logs --resource dup-1 --state creation` is executed, the TestManager has no idea where the resource CRD for that resource is, so the first API call will be needed to retrieve the UID, and the second API call to retrieve the actual creation pod.
It definitely seems reasonable to add a function to `resource_client.rs` and have the truncation logic consolidated there, but I don't see a way around two API calls.

model/src/test_manager/manager_impl.rs
```diff
@@ -214,19 +214,42 @@ impl TestManager {
...
+                        label_selector: Some(format!(
+                            "job-name={}{}-{}",
+                            resource_name,
+                            resource_object
+                                .metadata
+                                .uid
+                                .as_ref()
+                                .unwrap_or(&"".to_string()),
+                            suffix
+                        )),
```
ecpullen: This would become:
```
label_selector: Some(format!(
                             "job-name={}",
                             resource_crd.job_name(ResourceState::Creation))),
```

model/src/test_manager/manager_impl.rs
ecpullen: You would get the resource first.
`let resource = resource_client.get(my_resource)`
Then get the pod using `format!("job-name={}", resource.job_name(ResourceState::Create))`

mjsterckx: I addressed these changes in the new commit. Truncation now only happens in the `job_name` method in `resource.rs`.

ecpullen (approved):

etungsten (commented):
Changes look good to me.
The commit's subject is currently too long and it's getting truncated.

See [this comment](https://github.com/bottlerocket-os/bottlerocket/pull/2219#issuecomment-1176887596) by Ben a while back giving advice for writing commit messages.


For this particular change, maybe something like:

```
resource_controller: fix pod search filter

Fixed the 'resource_pod' method to search for the ...
```

etungsten (approved):

PR #526, opened 2022-08-18
test_controller: send error taskstate to test agent

### LINKED ISSUE #282: Agent Errors Might not be updating the CRD Status
> the status command only knows what the test crd and the controller know, I can change it to also look at the pod status if we want.

No, I think we should open a bug if this is reproducable. The CRD should have been set to an error state by the agent. It might be a bug with either the agent library of the test agent implementation.

_Originally posted by @webern in https://github.com/bottlerocket-os/bottlerocket-test-system/issues/275#issuecomment-1029395149_

### LINKED ISSUE #506: controller: Set test to errored state if config types are not correct
If the example test agents hello count is set to a string instead of a number, the test will remain in the `waiting` state.
```
$ cli status
 NAME                    TYPE                 STATE                 PASSED            SKIPPED            FAILED          
 duplicator              Resource             completed                                                                  
 hello-bones             Test                 waiting  
 ```

mjsterckx:
Modified controller to send Error TaskState to Test Agent in case of bad job state

**Issue number:**

Closes #506 

**Description of changes:**

In case of test job failure (for example, when an argument is given of the wrong type, like String instead of int), the controller would correctly set the JobFailure Error state but the TaskState of the Test was never updated. This would result in a perpetual "waiting" state, so the user would not know what was going on unless they looked at the test logs. This PR adds a function to the controller to set the TaskState of the test to TaskState::Error and give it the error message as displayed by the controller based on the bad job state. When using the `cli status` command, the state of the test now says `error` instead of `waiting` if a bad job state happens.

**Testing done:**

- Ran the `cli status` command for a test file that contained all arguments of the correct type (State is "passed")
- Ran the `cli status` command for a test file that contained an argument of the wrong type (State is "error")
- Ran the `cli status` command for a test file that takes an argument from a resource file with the correct type (State is "passed")
- Ran the `cli status` command for a test file that takes an argument from a resource file with the wrong type (State is "error")
- Ran the `cli status` command for a test file that cannot pull the image (State is "error" after the JobTimeout occurs)

**Terms of contribution:**

### DIFF
--- a/controller/src/test_controller/action.rs
+++ b/controller/src/test_controller/action.rs
@@ -106,7 +106,7 @@ pub(super) async fn determine_delete_action(t: &TestInterface) -> Result<Action>
     } else if t.test().has_finalizer(FINALIZER_MAIN) {
         Ok(Action::RemoveMainFinalizer)
     } else {
-        handle_bad_job_state(t, ErrorState::Zombie).await
+        Ok(Action::Error(ErrorState::Zombie))
     }
 }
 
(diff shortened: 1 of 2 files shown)


ecpullen:
Can you also check what happens if image pull fails? Currently the status is not updated correctly and I'm wondering if this fixes it as well.

ecpullen (approved):
Great work.

stmcginnis (approved):
Looks good to me.

Issue #417, opened 2022-05-19
minimal set of iam permissions


webern:
We need to come up with the minimal set of IAM permissions that are required to run our test and resource agents on EC2/EKS nodes.
Note that in #379 we want to give EKS and ECS providers a way to create clusters with create role or create instance profile permissions. So our minimal set of permissions should not include those (possibly with an option to add those in for the local developer workflow use case).

PR #527, opened 2022-08-19
Document minimum set of EKS IAM permissions

stmcginnis:
**Issue number:**

Partial: #417 

**Description of changes:**

This adds documentation on the minimum IAM permissions when running against an EKS cluster.
Additional work still needs to be done to determine the minimum permissions when working with ECS.

This also reorganizes some of our documentation now that we are getting more in the `bottlerocket-test-system` repo.
Some content is moved out of the main `README.md` into a dedicated `docs/` folder to keep the docs better organized and easier to discover.

**Testing done:**

Viewed rendered versions of docs.

**Terms of contribution:**


ecpullen (approved):
Looks great!

docs/IAMPERMISSIONS.md
```diff
@@ -0,0 +1,249 @@
+# Limiting IAM Permissions
+
+It is possible to define a role limiting IAM permissions when deploying to Amazon EKS or ECS.
```
etungsten: I don't think TestSys can be deployed to ECS.
```suggestion
It is possible to define a role limiting IAM permissions when deploying TestSys to Amazon EKS.
```

docs/IAMPERMISSIONS.md
```diff
@@ -0,0 +1,249 @@
...
+It is a best practice to use roles providing the minimum required permissions.
+The settings below have been found to be a good basis for assigning these minimum permissions when running `testsys` tests.
+
+## EKS
+
+The following is a minimal set of permissions when deploying to an EKS cluster:
```
etungsten, line 9: This seems like a lot just to get TestSys running in an EKS cluster.

If I understand this correctly, this seems like the IAM permissions necessary to _create_ an EKS cluster. In my mind, that seems to be out of the TestSys scope, because TestSys can be deployed to any K8s cluster and it doesn't need to be an EKS cluster.

I think what would be interesting is to provide set of minimum IAM permissions for the worker nodes' instance profile or things a user would have to do to run some of the common tests TestSys offers; Specifically, something like steps to allow test agents to be able to dispatch K8s conformance tests targeting external EKS clusters.

docs/QUICKSTART.md
```diff
@@ -0,0 +1,145 @@
...
+ "secret-access-key=$(aws configure get default.aws_secret_access_key)"
+```
+
+Now we are ready to run a Bottlerocket test.
+We get the latest AMI ID.
+Then we pass it to TestSys which will create an EKS cluster, launch Bottlerocket nodes and run a Sonobuoy test in 'quick' mode.
+
+**Caution**: The command below specifies `never` as the `--cluster-destruction-policy`.
+This is because creating a cluster takes a long time, and we might want to re-use it.
+To delete the cluster manually, use `eksctl delete cluster "external-cluster"` or delete the relevant CloudFormation stacks.
```
etungsten, line 82: We have instructions to delete clusters with `eksctl` here but no where before this do we provide instructions on creating the cluster via `eksctl`, we only have instructions on how to create the cluster locally with `kind`.

etungsten (changes requested):

docs/IAMPERMISSIONS.md
stmcginnis: Maybe reword this to something like "when creating and deploying to an EKS cluster"?

Today if someone follows our quickstart docs, they do need the permissions to be able to create the cluster. It's possible to use an existing one, but for someone new coming in and following those docs, they do need these permissions.

Maybe break out into a separate section that calls this out?

docs/QUICKSTART.md
stmcginnis: I believe we are able to create the cluster with `testsys`, but as far as I know we don't have anything built in to perform the clean up. So we do need to perform the clean up manually with `eksctl` or go through and delete things from the console.

stmcginnis: Also to point out - this is just moved out of the base README file, so this is what we currently have published.

docs/IAMPERMISSIONS.md
stmcginnis: Reworded slightly different. Let me know if that's more clear or if it needs further work. Thanks!

stmcginnis:
Based on further discussion, converting this to a draft for now to just serve as a reference for future work. We want to rework how we're doing things with the quickstart since we now have some duplication with content in the main `bottlerocket-os/bottlerocket` repo.

We'll like drop the quickstart here and just point to the other repo for details. With that, it may also make more sense for the IAM permissions to be documented there. We also would like to have permissions defined per test agent rather than one blanket example covering everything that could be done.

stmcginnis:
We're not going with this, and we can find it if needed by searching in the closed PRs.

PR #533, opened 2022-08-22
save results file in tar archive directory

vyaghras:
**Issue number:**

**Description of changes:**
This will help in saving file or directory in the output directory(tar archive)on test pod

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/agent/test-agent-cli/examples/example_test_agent_cli/example-test.sh
+++ b/agent/test-agent-cli/examples/example_test_agent_cli/example-test.sh
@@ -52,6 +52,13 @@ fi
 # Send the test results after consuming all the retries
 test-agent-cli send-result -o pass -p 1 -f 0 -s 0
 
+touch results.txt
+touch results.yaml
+mkdir my_results
+mkdir finished
+# Save the test results as a file or directory
+test-agent-cli save-results -f results.txt -f results.yaml -d my_results -d finished
+
 # Mark the test as completed.
-test-agent-cli terminate --results-dir results_directory
+test-agent-cli terminate
 
--- a/agent/test-agent-cli/src/save_results.rs
+++ b/agent/test-agent-cli/src/save_results.rs
@@ -1,0 +1,58 @@
+use crate::error::{ArchiveSnafu, Result};
+use argh::FromArgs;
+use copy_dir::copy_dir;
+use log::error;
+use model::constants::TESTSYS_RESULTS_DIRECTORY;
+use snafu::ResultExt;
+use std::path::Path;
+use test_agent::DefaultClient;
+
+#[derive(Debug, FromArgs, PartialEq)]
+#[argh(
+    subcommand,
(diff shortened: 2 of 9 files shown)


agent/test-agent-cli/src/save_results.rs
```diff
@@ -0,0 +1,71 @@
...
+#[argh(
+    subcommand,
+    name = "save-results",
+    description = "save the results file or directory in tar archive"
+)]
+pub(crate) struct SaveResults {
+    #[argh(short = 'd', option, description = "directory to add in tar archive")]
+    directory: Option<String>,
+    #[argh(short = 'f', option, description = "file to add in tar archive")]
+    file: Option<String>,
```
ecpullen: I think these should both be vecs.

ecpullen (changes requested):
Good work.

agent/test-agent-cli/src/save_results.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        if self.directory.is_empty() && self.file.is_empty() {
+            error!("Invalid arguments were provided. One of `--file`, `--directory` must be used.");
+        }
+
+        Ok(())
+    }
+
+    async fn move_file<P, Q>(&self, from: P, to: Q) -> Result<()>
```
ecpullen: Where is `self` used? If it isn't it can be moved out of the impl.

vyaghras: To call that method as self.move_file, we have to pass self as a param else I will have to call it as save_results::SaveResults::move_file. Is there any other better way to solve this ?

agent/test-agent-cli/examples/example_test_agent_cli/example-test.sh
```diff
@@ -48,10 +48,16 @@ if ! test && [ "$retries" -gt 0 ]; then
         fi
     done
 fi
-
```
stmcginnis: This line separation should probably be kept.

agent/test-agent-cli/src/save_results.rs
stmcginnis: Since `move_file` has no real association with `SaveResults` other than that that's where we are using it, maybe it should just be moved out of the `SaveResults` implementation and just be its own standalone function that we can call.

stmcginnis (commented):
Looks good overall, just one minor comment to add if you do an update.

Will wait to approve to get @ecpullen's response on the `move_file` preference.

agent/test-agent-cli/src/save_results.rs
ecpullen: You should be able to move it out and call `move_file`. It shouldn't need the extra stuff.

stmcginnis (approved):
Looks good!

ecpullen (approved):

PR #534, opened 2022-08-23
cli: add describe command to display object YAML

### LINKED ISSUE #532: cli: Provide a command to get tests
We should have a command to get the yaml of a test/resource.

mjsterckx:
**Issue number:**

Closes #532 

**Description of changes:**

Added a new `describe` command. This command takes one of two options (`--test <test-name>` or `--resource <resource-name>`) and gets the testsys object with this name. If the object is found, it converts the CRD to YAML representation and displays this. If not, an error message is shown.

**Testing done:**

- `cli describe --test <test-name>` with a valid test name: YAML is displayed
- `cli describe --test <test-name>` with an invalid test name (including the name of an existing resource): `Unable to get test` is shown
- `cli describe --resource <resource-name>` with a valid resource name: YAML is displayed
- `cli describe --resource <resource-name>` with an invalid resource name (including the name of an existing test): `Unable to get resource` is shown

**Terms of contribution:**

### DIFF
--- a/cli/src/describe.rs
+++ b/cli/src/describe.rs
@@ -7,11 +7,11 @@ use model::CrdExt;
 /// Retrieve the YAML description of a test or resource.
 #[derive(Debug, Parser)]
 pub(crate) struct Describe {
-    /// The name of the test we want the description from.
+    /// The name of the test to return the description from.
     #[clap(long, conflicts_with = "resource")]
     test: Option<String>,
 
-    /// The name of the resource we want the description from.
+    /// The name of the resource to return the description from.
     #[clap(long, conflicts_with = "test")]
     resource: Option<String>,
 }
--- a/cli/src/describe.rs
+++ b/cli/src/describe.rs
@@ -1,6 +1,5 @@
-use anyhow::{Context, Error, Result};
+use anyhow::{Error, Result};
 use clap::Parser;
-use futures::StreamExt;
 use model::test_manager::TestManager;
 
 /// Retrieve the YAML description of a test or resource.
@@ -19,16 +18,12 @@ impl Describe {
     pub(crate) async fn run(self, client: TestManager) -> Result<()> {
         match (self.test, self.resource) {
             (Some(test), None) => {
-                let mut logs = client.describe_test(test).await.context("Unable to get test description.")?;
-                while let Some(line) = logs.next().await {
-                    print!("{}", String::from_utf8_lossy(&line.context("Unable to read line")?));
-                }
+                let description = client.describe_test(test).await?;
+                println!("{}", description);
             }
             (None, Some(resource)) => {
-                let mut logs = client.describe_resource(resource).await.context("Unable to get resource description.")?;
-                while let Some(line) = logs.next().await {
-                    print!("{}", String::from_utf8_lossy(&line.context("Unable to read line")?));
-                }
+                let description = client.describe_resource(resource).await?;
+                println!("{}", description);
             }
             _ => return Err(Error::msg("Invalid arguments were provided. Exactly one of `--test` and `--resource` must be used.")),
         };
--- a/cli/src/describe.rs
+++ b/cli/src/describe.rs
@@ -1,0 +1,37 @@
+use anyhow::{Context, Error, Result};
+use clap::Parser;
+use futures::StreamExt;
+use model::test_manager::TestManager;
+
+/// Retrieve the YAML description of a test or resource.
+#[derive(Debug, Parser)]
+pub(crate) struct Describe {
+    /// The name of the test we want the description from.
+    #[clap(long, conflicts_with = "resource")]
+    test: Option<String>,
+
+    /// The name of the resource we want the description from.
+    #[clap(long, conflicts_with = "test")]
+    resource: Option<String>,
+}
+
+impl Describe {
+    pub(crate) async fn run(self, client: TestManager) -> Result<()> {
+        match (self.test, self.resource) {
+            (Some(test), None) => {
+                let mut logs = client.describe_test(test).await.context("Unable to get test description.")?;
+                while let Some(line) = logs.next().await {
+                    print!("{}", String::from_utf8_lossy(&line.context("Unable to read line")?));
+                }
+            }
+            (None, Some(resource)) => {
+                let mut logs = client.describe_resource(resource).await.context("Unable to get resource description.")?;
+                while let Some(line) = logs.next().await {
+                    print!("{}", String::from_utf8_lossy(&line.context("Unable to read line")?));
+                }
+            }
+            _ => return Err(Error::msg("Invalid arguments were provided. Exactly one of `--test` and `--resource` must be used.")),
+        };
+        Ok(())
+    }
+}
--- a/cli/src/describe.rs
+++ b/cli/src/describe.rs
@@ -1,6 +1,8 @@
 use anyhow::{Error, Result};
 use clap::Parser;
+use model::clients::CrdClient;
 use model::test_manager::TestManager;
+use model::CrdExt;
 
 /// Retrieve the YAML description of a test or resource.
 #[derive(Debug, Parser)]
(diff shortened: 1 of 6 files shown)


cli/src/describe.rs
```diff
@@ -0,0 +1,32 @@
...
+    /// The name of the resource we want the description from.
+    #[clap(long, conflicts_with = "test")]
+    resource: Option<String>,
+}
+
+impl Describe {
+    pub(crate) async fn run(self, client: TestManager) -> Result<()> {
+        match (self.test, self.resource) {
+            (Some(test), None) => {
+                let description = client.describe_test(test).await?;
```
ecpullen: We can get the test with 
`client.test_client().get(<test name>)`
and convert to string here.

ecpullen (changes requested):

cli/src/describe.rs
mjsterckx: Another option would be to add a `to_string` or `to_yaml` method to `crd_ext.rs` and call `client.list(&SelectionParams::Name(CrdName::Test(<test name>))).first().to_string()` from `describe.rs`. It doesn't look as clean, but I'm not sure if the CLI should know about/interact with `TestClient` (seeing as every other command only interacts with `TestManager` as well). If this is not a concern, however, I'd be happy to implement the suggested changes.

ecpullen: In the case of getting a single item. We should use `get` instead of `list`. When `TestManager` was created it was decided that methods from existing clients would not be duplicated i.e `test_manager::get_test()` because we already have `test_client::get()` and that a user would use the manager to get the client and then use the client to get the object (the logic behind that decision may need to be reconsidered). I think adding a function to `crd_ext` to get the yaml form of the object is a good start.

mjsterckx: I agree that using `list` is not great for a single item. I do think that there should be a `get_test()` method in `test_manager`, but that is up for discussion. For now, I can add the function to `crd_ext` and have `describe` get and print the CRD via the `test_client`.

jpmcb (commented):
This is mostly me musing on optimal CLI user experiences, but the UX of `describe` having a _flag_ that indicates which CRD is to be described feels a bit weird:

```
cli describe --<CRD> <name>
```

**Flags** are best reserved for some kind of "adjective" where they further modify the command or the subject of the command. For example:

```
APPNAME VERB NOUN --ADJECTIVE

or in other words

APPNAME COMMAND ARG --FLAG
```

The best CLI experiences flow like _written or spoken sentences_ and in this case, we are making the "noun" of the command the modifier which seems off.

---

A few proposed other workflows:

#### 1. (My preference) Use a `--name \ -n` flag instead of an `--<CRD>` flag

```sh
cli describe <CRD> --name <name>
cli describe <CRD> -n <name>
```

Examples:
```
cli describe test --name foo-test
cli describe resource -n bar-resource
```

This is a similar user experience to what `kind` does:
```sh
kind get cluster --name my--kind-cluster
```

#### 2. Use multiple modifier flags for both resource type and resource name. `--type / -t` for type of CRD being queried and `--name / -n` for the name of that resource.

```sh
cli describe --type <CRD> --name <name>
cli describe --t <CRD> -n <name>
```

Examples:
```
cli describe --type test --name foo-test
cli describe --t resource -n bar-resource
```

#### 3. Use a straight "commands" based approach:

```
cli describe <CRD> <name>
```

Examples being:
```
cli describe test foo-test
cli describe resource bar-resource
```

This isn't the best however since it creates additional "name" _subcommands_ on the existing `test` and `resource` subcommands creating possible confusion on _what_ is being modified by the last part of the command.

#### 4. Leave as is for now, but track overhauling some of these user experiences

I realize that this PR follows the current precedence for the user experience in our CLIs currently and maybe it's worth punting on this for now and tracking this in some kind of larger overhaul to the test sys CLI user experience.

Thoughts?

mjsterckx:
> Thoughts?

I agree with the reasoning, and my preference would be option 3 for the "cleanest" look though it might also bring some confusion. However, I decided to follow the structure used by the `cli logs` command, since `describe` is somewhat similar. It would be a good idea to take another look at how the commands are structured, but I think it should be part of a new issue/PR and not necessarily a concern for this one.

model/src/resource.rs
```diff
@@ -208,6 +208,10 @@ impl CrdExt for Resource {
     fn object_meta(&self) -> &ObjectMeta {
         self.meta()
     }
+
+    fn to_yaml(&self) -> Result<String, serde_yaml::Error> {
+        serde_yaml::to_string(self)
+    }
```
ecpullen: You should be able to move this to the body of to_yaml in `crd_ext`.

mjsterckx: That was my initial thought, but using `serde_yaml::to_string(self)` in `crd_ext` gives an error that `self` does not implement `Serialize`. Since `CrdExt` is a trait and not a struct, I wasn't sure if I could even make it derive `Serialize`.

ecpullen: You can define the trait as `pub trait CrdExt: Serialize` to force everything that uses the trait to be serializable.

ecpullen (approved):
LGTM

cli/src/describe.rs
```diff
@@ -0,0 +1,34 @@
+use anyhow::{Error, Result};
+use clap::Parser;
+use model::clients::CrdClient;
+use model::test_manager::TestManager;
+use model::CrdExt;
+
+/// Retrieve the YAML description of a test or resource.
+#[derive(Debug, Parser)]
+pub(crate) struct Describe {
+    /// The name of the test we want the description from.
```
jpmcb: nit: The cli saying `we` in the help messaging is a bit odd. Should be more of a "statement":

```
The name of the test to return the yaml for
```

cli/src/describe.rs
```diff
@@ -0,0 +1,34 @@
...
+    #[clap(long, conflicts_with = "resource")]
+    test: Option<String>,
+
+    /// The name of the resource we want the description from.
```
jpmcb: nit: Same here:

```
The name of the resource to return the yaml for
```

jpmcb (approved):
Looking good!

```
❯ cli describe --test hello-bones
---
apiVersion: testsys.bottlerocket.aws/v1
kind: Test
metadata:
  creationTimestamp: "2022-08-29T16:57:35Z"
  finalizers:
    - testsys.bottlerocket.aws/controlled
    - testsys.bottlerocket.aws/test-job
  generation: 1
  managedFields:
    - apiVersion: testsys.bottlerocket.aws/v1
      fieldsType: FieldsV1
      fieldsV1:
        "f:metadata":
          "f:finalizers":
            ".": {}
            "v:\"testsys.bottlerocket.aws/controlled\"": {}
            "v:\"testsys.bottlerocket.aws/test-job\"": {}
        "f:spec":
          ".": {}
          "f:agent":
            ".": {}
            "f:capabilities": {}
            "f:configuration":
              ".": {}
              "f:helloCount": {}
              "f:helloDurationMilliseconds": {}
              "f:mode": {}
              "f:person": {}
            "f:image": {}
            "f:keepRunning": {}
            "f:name": {}
            "f:pullSecret": {}
            "f:secrets": {}
            "f:timeout": {}
          "f:dependsOn": {}
          "f:resources": {}
          "f:retries": {}
      manager: unknown
      operation: Update
      time: "2022-08-29T16:57:35Z"
    - apiVersion: testsys.bottlerocket.aws/v1
      fieldsType: FieldsV1
      fieldsV1:
        "f:status":
          ".": {}
          "f:agent":
            ".": {}
            "f:error": {}
            "f:results": {}
            "f:taskState": {}
          "f:controller":
            ".": {}
            "f:resourceError": {}
      manager: unknown
      operation: Update
      time: "2022-08-29T16:57:35Z"
  name: hello-bones
  namespace: testsys-bottlerocket-aws
  resourceVersion: "3626"
  uid: 462e4266-0b73-4ee9-9830-b3c983352576
spec:
  resources: []
  dependsOn: []
  agent:
    name: hello-agent
    image: "example-test-agent:demo"
    pullSecret: ~
    keepRunning: false
    timeout: ~
    configuration:
      helloCount: 3
      helloDurationMilliseconds: 500
      mode: Fast
      person: Bones the Cat
    secrets: ~
    capabilities: ~
  retries: ~
status:
  controller:
    resourceError: ~
  agent:
    taskState: completed
    error: ~
    results:
      - outcome: pass
        numPassed: 0
        numFailed: 0
        numSkipped: 0
        otherInfo: ~
```

PR #536, opened 2022-08-23
agent: Unify agent  logic to crate agent-utils

ecpullen:
**Issue number:**
N/A

**Description of changes:**

This change moves commonly used functions of bottlerocket agents to their own crate that can be used by other agents.

**Testing done:**

**Terms of contribution:**


bottlerocket/agents/src/error.rs
```diff
@@ -184,4 +153,8 @@ pub enum Error {
 
     #[snafu(display("The task definition is missing"))]
     TaskDefinitionMissing,
+
+    #[snafu(context(false))]
+    #[snafu(display("{}", source))]
+    Utils { source: agent_utils::Error },
```
etungsten, line 159: I can't seem to find any references to this enum variant. Can we remove this?

etungsten (approved):

bottlerocket/agents/src/error.rs
ecpullen: That is a wrapper for any error that happened during a utils call. It is used, but snafu automatically converts it from utils::Error to bottlerocket_agents::Error.

etungsten: I see. `#[snafu(context(false))]` let's us not have to explicitly call `.context(UtilsSnafu)`. That's neat.

stmcginnis (approved):
lgtm

PR #537, opened 2022-08-24
agent: Custom derive to create builder for configs

ecpullen:
**Issue number:**
N/A

**Description of changes:**
In order to create an EC2 resource, the user needs to create the config, change it to a map, and give the templated string where it needs to go.
```
 fn ec2_resource(
        &self,
        name: &str,
        secrets: Option<BTreeMap<String, SecretName>>,
        cluster_resource_name: &str,
    ) -> Result<Resource> {
        let mut ec2_config = Ec2Config {
            node_ami: self.ami.clone(),
            // TODO - configurable
            instance_count: Some(2),
            instance_type: self.instance_type.clone(),
            cluster_name: self.cluster_name.clone(),
            region: self.region.clone(),
            instance_profile_arn: format!("${{{}.iamInstanceProfileArn}}", cluster_resource_name),
            subnet_id: format!("${{{}.privateSubnetId}}", cluster_resource_name),
            cluster_type: ClusterType::Eks,
            endpoint: Some(format!("${{{}.endpoint}}", cluster_resource_name)),
            certificate: Some(format!("${{{}.certificate}}", cluster_resource_name)),
            cluster_dns_ip: Some(format!("${{{}.clusterDnsIp}}", cluster_resource_name)),
            security_groups: vec![],
            assume_role: self.assume_role.clone(),
        }
        .into_map()
        .context(error::ConfigMapSnafu)?;

        // TODO - we have change the raw map to reference/template a non string field.
        let previous_value = ec2_config.insert(
            "securityGroups".to_owned(),
            Value::String(format!("${{{}.securityGroups}}", cluster_resource_name)),
        );
        if previous_value.is_none() {
            todo!("This is an error: fields in the Ec2Config struct have changed")
        }

        Ok(Resource {
            metadata: ObjectMeta {
                name: Some(name.to_string()),
                namespace: Some(NAMESPACE.into()),
                ..Default::default()
            },
            spec: ResourceSpec {
                depends_on: Some(vec![cluster_resource_name.to_owned()]),
                conflicts_with: None,
                agent: Agent {
                    name: "ec2-provider".to_string(),
                    image: self.ec2_provider_image.clone(),
                    pull_secret: self.ec2_provider_pull_secret.clone(),
                    keep_running: self.keep_instance_provider_running,
                    timeout: None,
                    configuration: Some(ec2_config),
                    secrets,
                    capabilities: None,
                },
                destruction_policy: DestructionPolicy::OnDeletion,
            },
            status: None,
        })
    }
```
This pr creates a builder type that can be used to create the entire crd from a config.
```
use configuration_build_derive::Configuration;
use serde::{Deserialize, Serialize};

#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default, Configuration)]
#[crd("Resource")]
struct Foo {
    bar: i32,
    biz: Option<String>,
}

fn main() {
    let a = Foo::builder()
        .biz(Some("b".to_string()))
        .bar(32)
        .biz(Some("Hello world".into()))
        .image("fake image")
        .build("test1")
        .unwrap();
    println!("{:?}", a);
    let b = Foo::builder()
        .bar_template("resource", "field")
        .biz_template("resource", "biz")
        .image("super secret image")
        .build("test1")
        .unwrap();
    println!("{:?}", b);
}
```
This creates 2 resources with `Foo` as the configuration. 

The EC2 example above would be reduced to:
```
fn ec2_resource(
        &self,
        name: &str,
        secrets: Option<BTreeMap<String, SecretName>>,
        cluster_resource_name: &str,
    ) -> Result<Resource> {
        let mut ec2_config = Ec2Config::builder().
          node_ami(self.ami.clone()).
          instance_count(Some(2)).
          instance_type(self.instance_type.clone()).
          cluster_name(self.cluster_name.clone()).
          region(self.region.clone()).
          instance_profile_arn_template(cluster_resource_name, "iamInstanceProfileArn").
          subnet_id_template(cluster_resource_name, "privateSubnetId").
          cluster_type(ClusterType::Eks).
          endpoint_template(cluster_resource_name, "endpoint").
          certificate_template(cluster_resource_name, "certificate").
          cluster_dns_ip_template(cluster_resource_name, "clusterDnsIp").
          security_groups_template(cluster_resource_name, "securityGroups").
          assume_role(self.assume_role.clone()).
          image(self.ec2_provider_image).
          pull_secret(self.ec2_provider_pull_secret).
          keep_running(self.keep_instance_provider_running).
          set_secrets(secrets).
          destruction_policy(DestructionPolicy::OnDeletion).
          build(name).context(error::CrdCreationSnafu)
}
```
**Testing done:**

Tested integration with sonobuoy config, but waiting to integrate macro throughout code base for additional pr.

**Terms of contribution:**


agent/configuration-build-derive/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+proc-macro2 = "1.0"
+serde = "1"
+
+[lib]
+proc-macro = true
+
+[dev-dependencies]
+serde = "1"
+serde_json= "1"
+model = { path = "../../model"}
```
etungsten: nit: missing newline

etungsten, line 20: nit: missing end line.

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,431 @@
+use proc_macro2::{Ident, Span, TokenStream};
```
etungsten: A docstring at the top of this module or in `lib.rs` explaining what this is and how to use it would be very helpful.

etungsten (commented):
Haven't looked into the deeper details yet, will do that soon. Just wanted to put out these comments first.

ecpullen:
^
* Added new line
* Added doc string to `lib.rs`
* Changed input for value setters to take `Into<#ty>` instead of `#ty` so that for an Option field, either `Some(value)` or just (value) can be used.

ecpullen:
^
* Added new line
* Added doc string to `lib.rs`
* Changed input for value setters to take `Into<#ty>` instead of `#ty` so that for an Option field, either `Some(value)` or just (value) can be used.

agent/configuration-build-derive/src/lib.rs
```diff
@@ -0,0 +1,28 @@
+//! Create a builder for TestSys agent configurations
+//!
+//! Each field of the configuration is given a setter for a templated
+//! value, or a typed value.
```
etungsten: This is a great start. I was thinking more along the lines of maybe adding [doctests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) to illustrate how to use the builder. Basically you can take what you have in `tests/derive.rs` (or maybe a pared down version of it) and add it as doctest here.

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,433 @@
...
+                                retries: Some(self.retries.as_ref().cloned().unwrap_or(5)),
+                                agent: model::Agent {
+                                    name: "agent".to_string(),
+                                    image: self.image.as_ref().cloned().ok_or_else(|| "Image is required to build a test".to_string())?,
+                                    pull_secret: self.image_pull_secret.as_ref().cloned(),
+                                    keep_running: self.keep_running.as_ref().cloned().unwrap_or(true),
+                                    timeout: None,
+                                    configuration: Some(configuration),
+                                    secrets: Some(self.secrets.clone()),
+                                    capabilities: Some(self.capabilities.clone()),
```
etungsten: Maybe add `..Default::default()` at the end so we don't have to come back to this every time we decided to add a new optional field? Same for resource spec.

agent/configuration-build-derive/tests/derive.rs
```diff
@@ -0,0 +1,47 @@
...
+
+#[test]
+fn test_derive() {
+    SampleTest::builder()
+        .biz(Some("b".to_string()))
+        .bar(32)
+        .biz(Some("Hello world".into()))
+        .image("fake image")
+        .build("test1")
+        .unwrap();
```
etungsten: Can we also try out some of the `set_*` functions? like `set_depends_on`, etc

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,433 @@
...
+                    where
+                    S1: Into<String> {
+                        self.depends_on.push(depends_on.into());
+                        self
+                    }
+
+                    pub fn set_depends_on(&mut self, depends_on: Option<Vec<String>>) -> &mut Self {
+                        self.depends_on = depends_on.unwrap_or_default();
+                        self
+                    }
```
etungsten: These `set_...` all look almost the same to me. Why do we take options for the argument and not just the type itself with `fn set_<field>` and we take the type itself and wrap it in an `Some` for `fn <field>`?

I wonder if there's a way to use `proc_macro` to generate and append these `set_*` setters by iterating over all the fields in the struct. So maybe something like (excuse the ugly formatting):
```
fn gen_set(crd: DataStruct) -> TokenStream {
    let mut set_stream = TokenStream2::default();
    for field in crd.fields {
        let fname=field.name;
        let ftype=field.type;
        set_stream.extend(
           quote! { 
                 fn set_#fname<S1>(&mut self, #fname: #ftype) -> &mut Self
                 where .... { 
                    self.#fname = #fname;
                    self
                 }
           }
       ) 
    }
    quote!{
        #set_stream
    }
}

...
impl #build_ident{
    #gen_set(#build_ident)
}
```

That way we wouldn't need to edit this whenever add a new optional field to the test or resource CRD.

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,433 @@
...
+
+                    #fns
+
+                    pub fn depends_on<S1>(&mut self, depends_on: S1) -> &mut Self
+                    where
+                    S1: Into<String> {
+                        self.depends_on.push(depends_on.into());
+                        self
+                    }
```
etungsten: What's the reasoning for having both `fn set_<field>` and `fn <field>`? Can we just stick with one? It seems like `fn <field>` is for adding one more to the collection and `fn set_<field>` is setting the entire collection. Do we need both?

ecpullen: I think it's important to have both. `set_` is not as clean looking as the other ones, but if a user is reusing the builder to create several similar test/resources a `set_` function is the only way to clear out the value.

agent/configuration-build-derive/src/derive.rs
ecpullen: ~~I don't think that would work for adding items to a vec. For instance when creating a set of migrations tests, we can create the initial test from a builder named `test_builder` . Then simply call `test_builder.depends_on("<INITIAL-TEST-NAME>").depends_on("<UPGRADE-TEST-NAME>").build()` to create the second test.~~
I didn't read close enough. Would we just ignore `name`, and `configuration` in the for loop?

agent/configuration-build-derive/src/derive.rs
ecpullen: I think that makes sense. We just may miss out on some updated agent features.

agent/configuration-build-derive/src/derive.rs
etungsten: I think it's probably better to just be more explicit and list out the setters as they are now.

ecpullen:
^ Adds unit testing and integrates the macro with `testsys run aws-k8s` command to see the improvements.

ecpullen:
^ Clippy fixes

stmcginnis (approved):
Looks like still a couple comments to address from @etungsten, but other than those, everything looks good to me.

ecpullen:
^ Adds default to `build` function, adds example doc test for a test crd.

ecpullen:
^ Fix merge conflicts.

agent/configuration-build-derive/tests/derive.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default, Configuration)]
+#[crd("Resource")]
+struct SampleResource {
+    bar: i32,
+    biz: Option<String>,
+}
+
+#[test]
+fn test_derive() {
```
etungsten: somewhat of a nit:
I think it's probably clearer to separate this unit test into multiple ones. Each unit test would be specific to a `::builder().build()` invocation and its subsequent checks.

etungsten (approved):

etungsten (approved):

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,434 @@
...
+                .unwrap_or(false)
+        });
+        let field_name = match field.ident.as_ref() {
+            Some(ident) => ident.to_string(),
+            None => return None,
+        };
+        let field_ident = Ident::new(&field_name, Span::call_site());
+        let template_field_name = format!("{}{}", field_name, "_template");
+        let template_ident = Ident::new(&template_field_name, Span::call_site());
+        let ty = field.ty.clone();
```
webern: It's not your fault that the library you're using has a variable named `yt`, but can we do better? Is this `field_type`?

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,434 @@
...
+                            depends_on: Some(self.depends_on.clone()),
+                            agent: model::Agent {
+                                name: "agent".to_string(),
+                                image: self.image.as_ref().cloned().ok_or_else(|| "Image is required to build a test".to_string())?,
+                                pull_secret: self.image_pull_secret.as_ref().cloned(),
+                                keep_running: self.keep_running.as_ref().cloned().unwrap_or(true),
+                                configuration: Some(configuration),
+                                secrets: Some(self.secrets.clone()),
+                                capabilities: Some(self.capabilities.clone()),
+                                ..Default::default()
```
webern: Same here, better to fail compilation than silently have a field we confusingly can't use the builder for.

agent/configuration-build-derive/src/derive.rs
```diff
@@ -0,0 +1,434 @@
...
+use syn::{self, Data, LitStr};
+
+extern crate quote;
+
+pub(crate) fn impl_configuration(ast: &syn::DeriveInput) -> TokenStream {
+    let builder_name = format!("{}{}", &ast.ident.to_string(), "Builder");
+    let builder_ident = Ident::new(&builder_name, Span::call_site());
+    let ident = &ast.ident;
+    quote! {
+       impl model::Configuration for #ident{}
```
webern: I think it would be better to have separate derive macros for `Configuration` and `Builder`. If we're doing that then we might also want to have `TestBuilder` and `ResourceBuilder` be separate.

My thinking behind this is:
* Seeing `#[derive(Configuration)]` implies to me that I am implementing the `Configuration` trait. It doesn't imply to me that extra structs are being created.
* I might, for some reason, not want to have the builder generated (maybe it conflicts with the name of a struct who's name I cannot change or whatever).
* I think it is easier when reading the code to understand that I am both implementing Configuration and creating a builder if I see this `#[derive(Configuration, Builder)]`

agent/configuration-build-derive/src/derive.rs
webern: I actually think it's better to have a compilation error so that I understand that there are extra steps involved with adding a field to `Agent`.

In otherwords, if I add `foo: String` to the `Agent` struct, then someone later would be very confused as to why the generated builder doesn't allow you to set `foo`. To avoid that, I would rather have the compiler fail during the derive macro so that I am forced to add it to this builder.

webern (changes requested):
This is pretty cool. I think it will be easier when reading the `#derive` lines if we separate the `Configuration` trait and `Builder` struct into separate `derive`s.

agent/configuration-build-derive/src/derive.rs
ecpullen: `ty` is used in almost every macro I've read. It seems like we shouldn't stray from common conventions.

webern: I don't like it, but it's definitely not worth blocking over. Just because somebody did a bad thing and used a variable named `ty` doesn't mean I should. If everyone jumped off a cliff...

ecpullen:
^ split macro into `Configuration` and `Builder`

ecpullen:
^ Code cleanup

ecpullen:
^ Fixed doc test issue.

webern (approved):
This is cool.

I might have opted to put both derive macros into a single `derive` crate, but we can consolidate if we find ourselves creating a third crate.

ecpullen:
^ Fixes merge conflicts

PR #553, opened 2022-09-06
test-agent: Wait for terminate before sending done

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Before this change, a test was marked as complete before starting its termination. This can cause issues with automatic resource destruction. Instead, the test will terminate before the `Complete` state is sent.

**Testing done:**

Updated sonobuoy image was used to test conformance and migration testing on the same cluster and tests passed successfully.

**Terms of contribution:**

### DIFF
--- a/agent/test-agent/src/agent.rs
+++ b/agent/test-agent/src/agent.rs
@@ -131,7 +131,7 @@ where
 
         if let Err(e) = self
             .client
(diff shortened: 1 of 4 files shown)


etungsten (approved):
LGTM assuming tests pass

jpmcb (approved):

Issue #379, opened 2022-04-20
allow cluster creation without iam create role permissions


webern:
IAM create role is a permission that we should be extremely protective of. We should make it possible to avoid assigning this role to testsys nodes and containers. There are some related roles that we probably want to avoid such as "attach role to instance profile"

This applies to
- [x] The EKS provider
- [x] The ECS provider

On the EKS side this is related to
- #225 because a configuration file would probably be used to pass the preexisting role
- #252 because it might also involve a configuration file
- #218 because we should provide instructions for the minimal set of permissions for various roles the system needs

There may be a more direct reference in eksctl documentation, but when I was searching I found an example here where it appears the role was created before the eksctl create cluster call:

https://reece.tech/posts/building-eks-clusters-using-eksctl/

A complication to consider: the EKS provider has logic for finding roles based on eksctl's naming convention.

ecpullen:
This was closed with #447.

PR #555, opened 2022-09-07
ecs-resource-agent: added field for existing profile

mjsterckx:
Added field for existing IAM profile to the CRD and code to check if the field contains a value: if so, retrieves the ARN of the submitted profile instead of the default one.

**Issue number:**

#379 

**Description of changes:**

Added a field to the ecs-resource-agent CRD where a user can enter the name of an existing IAM instance profile. This allows cluster creation without IAM role creation permissions. The ecs-provider code will check if the field is used, and if so, attempt to retrieve the ARN of the submitted profile instead of the default one. If this is unsuccessful (because, for example, the profile does not exist), the ECS creation throws an error.

**Testing done:**

- Ran ECS test without the field: everything ran exactly like before the changes
- Ran ECS test with a profile name that does not exist:
`Provider error: An error occurred but no resources were left behind, The iam instance profile name was not found.: An error left resources behind that can be destroyed, Unable to get instance profile.: NoSuchEntityException: Instance Profile does-not-exist cannot be found.`
- Ran ECS test with a profile name that does exist: everything ran exactly like before the changes

**Terms of contribution:**

### DIFF
--- a/bottlerocket/testsys/src/run_aws_ecs.rs
+++ b/bottlerocket/testsys/src/run_aws_ecs.rs
@@ -146,6 +146,11 @@ pub(crate) struct RunAwsEcs {
     /// The arn for the role that should be assumed by the agents.
     #[structopt(long)]
     assume_role: Option<String>,
+
+    /// The IAM instance profile name for the EC2 instances in the ECS cluster. If no value is
+    /// provided, then the ECS test agent will attempt to create an IAM instance profile.
+    #[structopt(long)]
+    iam_instance_profile_name: Option<String>,
 }
 
 impl RunAwsEcs {
@@ -326,6 +331,7 @@ impl RunAwsEcs {
                             region: Some(self.region.clone()),
                             vpc: self.vpc.clone(),
                             assume_role: self.assume_role.clone(),
+                            iam_instance_profile_name: self.iam_instance_profile_name.clone(),
                         }
                         .into_map()
                         .context(error::ConfigMapSnafu)?,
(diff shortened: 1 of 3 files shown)


bottlerocket/testsys/src/run_aws_ecs.rs
```diff
@@ -146,6 +146,10 @@ pub(crate) struct RunAwsEcs {
     /// The arn for the role that should be assumed by the agents.
     #[structopt(long)]
     assume_role: Option<String>,
+
+    /// The IAM instance profile name. If no value is provided the default name will be used.
```
etungsten: I think we can elaborate a bit more on this.
```suggestion
    /// The IAM instance profile name for the EC2 instances in the ECS cluster. If no value is provided, then the ECS test agent will attempt to create an IAM instance profile.
```

bottlerocket/types/src/agent_config.rs
```diff
@@ -212,6 +212,9 @@ pub struct EcsClusterConfig {
 
     /// The role that should be assumed when creating the ecs cluster.
     pub assume_role: Option<String>,
+
+    /// The IAM instance profile name. If no value is provided the default name will be used.
```
etungsten: Similar to above,
```suggestion
    /// The IAM instance profile name for the EC2 instances in the ECS cluster. If no value is provided, then the ECS test agent will attempt to create an IAM instance profile.
```

etungsten (approved):

ecpullen (approved):

PR #563, opened 2022-09-08
Actions: update docs directory path

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

We originally had a `design` directory that contained the design doc for testsys. We added more docs and restructured things, so `design` was renamed to `docs`. There is a GitHub Action configuration that tries to ignore this path to skip running tests if only docs have changed. This updates the config to use the correct current name.

**Testing done:**

Will need to verify with job triggers after this merges.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,13 +3,13 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'
-      - 'design/**'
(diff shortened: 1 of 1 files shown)


etungsten (approved):

jpmcb (approved):
Good catch! 🕺🏼

PR #564, opened 2022-09-09
utils: added code to check for optional session token

### LINKED ISSUE #562: utils: check for session_token in credentials
Currently `aws.rs` does not check for a session token in the provided credentials, which prevents temporary credentials from an assumed role from working.

mjsterckx:
**Issue number:**

Closes #562 

**Description of changes:**

Added code to `setup_resource_env` to check the provided secret for a session token, and set the environment variable if a token is present. The session token is needed when temporary credentials from an assumed user are used.

**Testing done:**

- Ran code to create cluster providing secret with assumed role and no session token: fails to create instances
- Ran code to create cluster providing secret with assumed role and incorrect session token: fails to create instances
- Ran code to create cluster providing secret with assumed role and correct session token: instances are created

**Terms of contribution:**

### DIFF
--- a/agent/utils/src/aws.rs
+++ b/agent/utils/src/aws.rs
@@ -271,6 +271,16 @@ where
         resources,
         "Could not convert secret-access-key to String",
     )?;
+    if let Some(token) = aws_secret.get("session-token") {
+        env::set_var(
+            "AWS_SESSION_TOKEN",
+            resource_agent::provider::IntoProviderError::context(
+                String::from_utf8(token.to_owned()),
+                resources,
+                "Could not convert session-token to String",
+            )?,
+        );
+    };
 
     env::set_var("AWS_ACCESS_KEY_ID", access_key_id);
     env::set_var("AWS_SECRET_ACCESS_KEY", secret_access_key);


agent/utils/src/aws.rs
```diff
@@ -271,6 +271,17 @@ where
...
+        Some(token) => env::set_var(
+            "AWS_SESSION_TOKEN",
+            resource_agent::provider::IntoProviderError::context(
+                String::from_utf8(token.to_owned()),
+                resources,
+                "Could not convert session-token to String",
+            )?,
+        ),
+        None => (),
+    };
```
etungsten: I think this can be better expressed as:
```suggestion
    if let Some(token) = aws_secret.get("session-token") {
        env::set_var(
            "AWS_SESSION_TOKEN",
            resource_agent::provider::IntoProviderError::context(
                String::from_utf8(token.to_owned()),
                resources,
                "Could not convert session-token to String",
            )?;
    }
```

ecpullen (approved):
LGTM

etungsten (approved):

PR #591, opened 2022-09-26
TestManager: Custom columns for status table

ecpullen:
**Issue number:**

A step towards https://github.com/bottlerocket-os/bottlerocket/issues/2422
After this pr is merged, `cargo make test` can add the build id as a label and use the new crd label function to write a custom column as `BUILD ID`.

**Description of changes:**

This pr allows users to add additional columns to the status table via `StatusSnapshot::new_column(<HEADER>, <FUNCTION TO GET VALUE>)`. For each crd, the function will be run and the response will be added as an additional column.

**Testing done:**
```rust
status.new_column("ARCH", |crd| crd.labels().get("testsys/arch").cloned())
```
Creates the usual status window with an additional column displaying the value for `testsys/arch` or nothing if the key is not set.

**Terms of contribution:**

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -24,6 +24,7 @@ pub use resource::{
     ResourceStatus,
 };
 use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
 pub use test::{
     AgentStatus, ControllerStatus, Outcome, Test, TestResults, TestSpec, TestStatus, TestUserState,
 };
@@ -69,6 +70,13 @@ impl Crd {
         match self {
             Self::Test(test) => test.metadata.name.to_owned(),
             Self::Resource(resource) => resource.metadata.name.to_owned(),
+        }
+    }
+
+    pub fn labels(&self) -> BTreeMap<String, String> {
+        match self {
+            Self::Test(test) => test.metadata.labels.to_owned().unwrap_or_default(),
+            Self::Resource(resource) => resource.metadata.labels.to_owned().unwrap_or_default(),
         }
     }
 }
--- a/model/src/test_manager/status.rs
+++ b/model/src/test_manager/status.rs
@@ -3,12 +3,16 @@ use k8s_openapi::api::core::v1::PodStatus;
 use kube::{core::object::HasStatus, ResourceExt};
 use serde::Serialize;
 use tabled::object::Segment;
-use tabled::{Alignment, MaxWidth, MinWidth, Modify, Style, Table, Tabled};
+use tabled::{
+    Alignment, Concat, Extract, MaxWidth, MinWidth, Modify, Style, Table, TableIteratorExt, Tabled,
+};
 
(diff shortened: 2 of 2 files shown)


stmcginnis (approved):

mjsterckx (approved):

model/src/lib.rs
```diff
@@ -71,6 +72,18 @@ impl Crd {
...
     }
+
+    pub fn labels(&self) -> BTreeMap<String, String> {
+        match self {
+            Self::Test(test) => test.metadata.labels.as_ref().cloned().unwrap_or_default(),
+            Self::Resource(resource) => resource
+                .metadata
+                .labels
+                .as_ref()
+                .cloned()
```
etungsten: Can `as_ref().clone()` be simplified to  `to_owned()`?
```suggestion
            Self::Test(test) => test.metadata.labels.to_owned().unwrap_or_default(),
            Self::Resource(resource) => resource
                .metadata
                .labels
                .to_owned()
```

model/src/test_manager/status.rs
```diff
@@ -86,6 +102,8 @@ impl std::fmt::Display for StatusSnapshot {
 
 impl From<&StatusSnapshot> for Table {
     fn from(status: &StatusSnapshot) -> Self {
+        let mut crds = status.crds.clone();
+        crds.sort_by_key(|crd| crd.name());
```
etungsten, line 108: With the additional complexity, some unit tests might be nice to have for this trait implementation.

etungsten:
Another side comment, to keep with our contribution guidelines, the PR branch should probably be created in your fork instead of the main project repository.

ecpullen:
^ Change `as_ref().cloned()` to `to_owned()`. Added an example for adding an additional column.

model/src/test_manager/status.rs
```diff
@@ -198,3 +245,22 @@ impl From<&Crd> for Vec<ResultRow> {
...
+            .field("header", &self.header)
+            .finish()
+    }
+}
+
+#[test]
+fn test_new_columns() {
+    // Add a new column to the status with the crd's name.
+    StatusSnapshot::new(None, Default::default()).new_column("extra column", |crd| crd.name());
+}
```
etungsten: This test would never fail so I'm not sure if this is useful. If we want to show an example for adding a new custom column, we can probably have it as a docstring for `StatusSnapshot`.

model/src/test_manager/status.rs
etungsten: I'm not too familiar with `tabled::Table` so maybe it's not easy to `asset_eq` to ensure the generated table is as expected in a unit test.

model/src/test_manager/status.rs
ecpullen: It would be tough to create an `assert` with the Tabled lib.

etungsten (approved):

PR #585, opened 2022-09-23
ec2_provider: try to create instances in any subnet in region

mjsterckx:
**Issue number:**

Closes #578 

**Description of changes:**

Instead of the CRDs containing a single subnet, they will now contain a vector of all of the subnets in a region. This allows the EC2 provider to try to run instances on different availability zones if some of them do not support the instance type. The EC2 provider will iterate over the subnets and try to run instances in each one. If the run_instances method is successful, the function continues with the expected behavior. If not, the function will show a warning and attempt the next subnet. If none of them worked, the function errors out just like before the changes.

**Testing done:**

- Ran instances with GPUs on us-west-2. Before the changes, the code would pick availability zone us-west-2d where GPU instances were not supported. With these changes, the following logs are shown:

```
[2022-09-22T23:05:04Z INFO  resource_agent::agent] Initializing Agent
[2022-09-22T23:05:04Z INFO  ec2_resource_agent::ec2_provider] Beginning instance creation with instance UUID: ******************************
[2022-09-22T23:05:04Z INFO  ec2_resource_agent::ec2_provider] Using instance type 'g2.2xlarge'
[2022-09-22T23:05:04Z INFO  ec2_resource_agent::ec2_provider] Creating 2 instance(s)
[2022-09-22T23:05:04Z INFO  ec2_resource_agent::ec2_provider] Starting instances in subnet subnet-d*******
[2022-09-22T23:05:05Z WARN  ec2_resource_agent::ec2_provider] An error occurred while trying to create instances using subnet subnet-d*******: Error { code: "Unsupported", message: "Your requested instance type (g2.2xlarge) is not supported in your requested Availability Zone (us-west-2d). Please retry your request by not specifying an Availability Zone or choosing us-west-2a, us-west-2b, us-west-2c." }
[2022-09-22T23:05:05Z INFO  ec2_resource_agent::ec2_provider] Starting instances in subnet subnet-c*******
[2022-09-22T23:05:07Z INFO  ec2_resource_agent::ec2_provider] Created instances using subnet subnet-c*******
[2022-09-22T23:05:07Z INFO  ec2_resource_agent::ec2_provider] Checking instance IDs
[2022-09-22T23:05:07Z INFO  ec2_resource_agent::ec2_provider] Waiting for instances to reach the running state
[2022-09-22T23:05:18Z INFO  ec2_resource_agent::ec2_provider] Done: instances {"i-********************", "i-*******************"} are running
```
- Ran `aws_k8s` which resulted in the following output:
```
[...]

Issue #578, opened 2022-09-19
ecs-provider/eks-provider: return all public subnet IDs


mjsterckx:
The ECS provider and EKS provider currently create a cluster and return the first public subnet ID to be used by the EC2 provider. This determines the availability zone where the EC2 provider can create instances. It is possible that certain instance types are not supported on an availability zone, but are supported on other availability zones in the region. The EC2 provider should not be limited to a single availability zone when trying to create instances, but instead iterate over all of the subnets returned by the ECS/EKS provider to see if the desired instance type is supported and available on any of them.

PR #602, opened 2022-10-11
ec2-resource: accept vector of instance types

### LINKED ISSUE #525: ec2-resource: Provide fallback methods for unavailable instance type
It is possible that the desired instance type will not be available. If that is the case, the user should have the ability of providing alternative instance types that can be used.

mjsterckx:
**Issue number:**

Closes #525 

**Description of changes:**

Changed the `ec2_resource_agent`'s `instance_type` which accepted a string to `instance_types` which accepts a vector of strings. A user can now submit multiple instance types and the EC2 resource agent will try to provision instances of the types in the list until the list is exhausted or instances were successfully created. In case of an error (such as UnsupportedException or InstanceCapacityException) the agent will move on to the next type in the list.

**Testing done:**

- Ran an ECS test with a file containing a vector of three instance types. The EC2 resource agent tried to reserve instances of the first type in each availability zone in the given region, but this resulted in an InstanceCapacityException for each zone. It then moved on to the second type, which returned an UnsupportedException in the first availability zone, but the instances of this second type were successfully created in the second availability zone.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
+++ b/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
@@ -118,86 +118,9 @@ impl Create for Ec2Creator {
         .await?;
         let ec2_client = aws_sdk_ec2::Client::new(&shared_config);
 
-        // Determine the instance type to use. If provided use that one. Otherwise, for `x86_64` use `m5.large`
-        // and for `aarch64` use `m6g.large`
-        let instance_type = if let Some(instance_type) = spec.configuration.instance_type {
-            instance_type
-        } else {
-            instance_type(&ec2_client, &spec.configuration.node_ami, &memo).await?
-        };
-        info!("Using instance type '{}'", instance_type);
+        let run_instance_result = run_instances(&spec, &ec2_client, instance_uuid, &memo).await?;
 
-        // Run the ec2 instances
-        let instance_count = spec
-            .configuration
-            .instance_count
-            .unwrap_or(DEFAULT_INSTANCE_COUNT);
-        info!("Creating {} instance(s)", instance_count);
-
-        let mut run_instance_result = None;
-        for subnet_id in &spec.configuration.subnet_ids {
-            let run_instances = ec2_client
-                .run_instances()
-                .min_count(instance_count)
-                .max_count(instance_count)
-                .subnet_id(subnet_id)
-                .set_security_group_ids(Some(spec.configuration.security_groups.clone()))
-                .image_id(&spec.configuration.node_ami)
-                .instance_type(InstanceType::from(instance_type.as_str()))
-                .tag_specifications(tag_specifications(
-                    &spec.configuration.cluster_type,
-                    &spec.configuration.cluster_name,
-                    &instance_uuid,
-                ))
-                .user_data(userdata(
-                    &spec.configuration.cluster_type,
-                    &spec.configuration.cluster_name,
-                    &spec.configuration.endpoint,
-                    &spec.configuration.certificate,
-                    &spec.configuration.cluster_dns_ip,
-                    &memo,
-                )?)
-                .iam_instance_profile(
-                    IamInstanceProfileSpecification::builder()
-                        .arn(&spec.configuration.instance_profile_arn)
-                        .build(),
-                );
-
-            info!("Starting instances in subnet {}", subnet_id);
-            let run_instance_output = tokio::time::timeout(
-                Duration::from_secs(360),
-                wait_for_successful_run_instances(&run_instances),
-            )
-            .await
-            .context(
-                Resources::Clear,
-                "Failed to run instances within the time limit",
-            )?;
-            match run_instance_output {
-                Ok(result) => {
-                    info!("Created instances using subnet {}", subnet_id);
-                    run_instance_result = Some(result);
-                    break;
-                }
-                Err(err) => {
-                    warn!(
-                        "An error occurred while trying to create instances using subnet {}: {}",
-                        subnet_id, err
-                    );
-                }
-            }
-        }
-
-        let result = match run_instance_result {
-            Some(result) => result,
-            None => {
-                return Err(ProviderError::new_with_context(
-                    Resources::Clear,
-                    "Failed to create instances in any of the provided subnets",
-                ))
-            }
-        };
-        let instances = result
+        let instances = run_instance_result
             .instances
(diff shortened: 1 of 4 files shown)


bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -120,70 +120,73 @@ impl Create for Ec2Creator {
...
+            vec![instance_type(&ec2_client, &spec.configuration.node_ami, &memo).await?]
         };
-        info!("Using instance type '{}'", instance_type);
-
-        // Run the ec2 instances
-        let instance_count = spec
-            .configuration
-            .instance_count
-            .unwrap_or(DEFAULT_INSTANCE_COUNT);
-        info!("Creating {} instance(s)", instance_count);
```
ecpullen: This should stay outside of the for loop.

bottlerocket/testsys/src/run_aws_ecs.rs
```diff
@@ -355,7 +355,10 @@ impl RunAwsEcs {
             node_ami: self.ami.clone(),
             // TODO - configurable
             instance_count: Some(2),
-            instance_type: self.instance_type.clone(),
+            instance_types: match self.instance_type.clone() {
+                Some(instance_type) => vec![instance_type],
+                None => vec![],
+            },
```
ecpullen: This may be cleaner: `self.instance_type.clone().iter().collect()` or even `self.instance_type.map(|instance_type| vec![instance_type.clone()]).unwrap_or_default()`.

bottlerocket/types/src/agent_config.rs
```diff
@@ -146,7 +146,7 @@ pub struct Ec2Config {
     /// The type of instance to spin up. m5.large is recommended for x86_64 and m6g.large is
     /// recommended for arm64 on eks. c3.large is recommended for ecs. If no value is provided
     /// the recommended type will be used.
-    pub instance_type: Option<String>,
+    pub instance_types: Vec<String>,
```
ecpullen, line 149: Could/should we except both ways to avoid breaking changes?
@webern @etungsten

webern: I wouldn't want to have two fields (`instance_type` and `instance_types`). It might be possible with something like:

`Either<Option<String>, Option<Vec<String>>`

But I think we might be trying to hard for backwards compatibility for a 0.0.2 versioned system.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
mjsterckx: Line 131-134 definitely, but I thought we could use some kind of message mentioning the specific instance type we are currently trying per iteration of the loop.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -120,70 +120,73 @@ impl Create for Ec2Creator {
...
-                    &spec.configuration.cluster_dns_ip,
-                    &memo,
-                )?)
-                .iam_instance_profile(
-                    IamInstanceProfileSpecification::builder()
-                        .arn(&spec.configuration.instance_profile_arn)
-                        .build(),
-                );
+        'outer: for instance_type in instance_types {
+            info!("Using instance type '{}'", instance_type);
```
etungsten: Maybe we can refactor this into a function and just return  the `instance_result` when we've successfully launched the instances instead of breaking loops awkwardly here.

bottlerocket/types/src/agent_config.rs
mjsterckx: Just to clarify: do the "thumbs up" mean that people are in favor of the `Either` option or that we shouldn't worry about backwards compatibility too much and can leave it as-is?

etungsten: Although unfortunate, I think what you have right now is fine.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
webern: I agree. The "inner loop" (for subnet id) could be done in a function which would improve readability. @ecpullen why do we loop over subnet ids?

mjsterckx: I'll put that inner loop in a different function. Some context for why we loop over subnet ids is in this PR: https://github.com/bottlerocket-os/bottlerocket-test-system/pull/585 and this issue: https://github.com/bottlerocket-os/bottlerocket-test-system/issues/578 . Essentially, it was possible that an instance type was not available on the randomly selected availability zone but it might be available on another one, so we try to create the instance on each AZ until one works.

webern: One possible issue with what we are doing here... we aren't checking the reason for the failure. So if I understand correctly, we might receive an error such as "Not Authorized" and we would go through all of this looping anyway, right? Or are we checking that the issue is actually a lack of capacity?

mjsterckx: We are not checking the specific type of error, since there is a multitude of errors that we should still loop for (InstanceCapacityException, UnsupportedException, etc.). Exiting the loop immediately upon return of a certain error is certainly something that can be added, but I don't know if that's a priority right now. Would love some opinions.

mjsterckx:
^ moved the for-loops to a different function to improve readability and eliminate `break`ing

etungsten (approved):

webern (approved):

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -373,7 +373,11 @@ impl RunAwsK8s {
             node_ami: self.ami.clone(),
             // TODO - configurable
             instance_count: Some(2),
-            instance_type: self.instance_type.clone(),
+            instance_types: self
+                .instance_type
+                .clone()
+                .map(|instance_type| vec![instance_type])
+                .unwrap_or_default(),
```
ecpullen, line 380: I think this will also work, `self.instance_type.to_owned().iter().collect()`

Not a blocker but you can switch it over if you like it better.

ecpullen (approved):
LGTM

PR #607, opened 2022-10-14
controller: replaced parse_duration package with custom parser

### LINKED ISSUE #596: Replace `parse_duration` package
`parse_duration` is no longer being maintained, a new library should be used instead.

mjsterckx:
**Issue number:**

Closes #596 

**Description of changes:**

The `parse_duration` package is no longer being maintained. Since we used the package for a very specific use-case (parsing the `timeout` string, which is already validated by regex), it was easier to remove the package and implement a short function to replace it. This function takes a string that conforms to `"^((([0-9]+)d)?(([0-9]+)h)?(([0-9]+)m)?(([0-9]+)s)?|\d+)$"` and converts it into a `Duration` object.

**Testing done:**

Created `Test` object with various valid values for `timeout` ("123", "1d2h3m4s", "500s", "3h100s", "20d", etc.) which all resulted in the correct `Duration`. When a value was used for `timeout` that didn't conform to the regex, the object could not be created (as is expected).

**Terms of contribution:**

### DIFF
--- a/controller/Cargo.toml
+++ b/controller/Cargo.toml
@@ -16,7 +16,6 @@ kube-runtime = "0.75"
 lazy_static = "1"
 log = "0.4"
 model = { version = "0.0.2", path = "../model" }
-parse_duration = "2.1"
 schemars = "=0.8.10"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/controller/src/constants.rs
+++ b/controller/src/constants.rs
@@ -1,5 +1,10 @@
+use crate::error::Result;
+use anyhow::Context;
 use kube_runtime::controller::Action;
+use std::collections::VecDeque;
 use std::time::Duration;
+
+const UNITS: [(char, u64); 3] = [('d', 86400), ('h', 3600), ('m', 60)];
 
 /// Tell the controller to reconcile the object again after some duration.
 pub(crate) fn requeue() -> Action {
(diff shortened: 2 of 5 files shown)


controller/src/constants.rs
```diff
@@ -15,3 +17,20 @@ pub(crate) fn requeue_slow() -> Action {
 pub(crate) fn no_requeue() -> Action {
     Action::await_change()
 }
+
+/// Parse a duration string into a Duration object.
+pub(crate) fn parse_duration(input: &str) -> Result<Duration, ParseIntError> {
+    let mut secs: u64 = 0;
+    let mut duration_string = input;
+    let units = vec![('d', 86400), ('h', 3600), ('m', 60)];
```
webern: Not super important but this could be a constant at the top of the file.

const UNITS: [(char, int32); 3] = [('d', 86400), ('h', 3600), ('m', 60)];

Seems like this file should be called `parse_duration.rs` or `parse.rs` or something like that.

webern (changes requested):
Nice. Please add unit tests.

controller/src/constants.rs
mjsterckx: I added the function to the `constants.rs` file because it seemed like that file contained functions used by all controllers. However, I can definitely move it to its own file if that would be more idiomatic.

webern: It's fine to leave it in constants.rs, probably better than adding a new file or renaming a file for one function that doesn't "fit"

mjsterckx:
Since the new function is only called with values that are previously checked with regex, running the function with a value that doesn't conform to the regex is undefined behavior. Should I "define" this behavior and add some unit tests for those cases, or only add tests for values that conform to the regex?

controller/Cargo.toml
```diff
@@ -16,7 +16,6 @@ kube-runtime = "0.75"
 lazy_static = "1"
 log = "0.4"
 model = { version = "0.0.2", path = "../model" }
-parse_duration = "2.1"
```
ecpullen, line 19: Shouldn't this also be removed from the `Cargo.lock`?

controller/src/constants.rs
```diff
@@ -15,3 +17,20 @@ pub(crate) fn requeue_slow() -> Action {
 pub(crate) fn no_requeue() -> Action {
     Action::await_change()
 }
+
+/// Parse a duration string into a Duration object.
```
ecpullen: nit: `Duration`

controller/src/constants.rs
```diff
@@ -15,3 +17,20 @@ pub(crate) fn requeue_slow() -> Action {
...
+pub(crate) fn parse_duration(input: &str) -> Result<Duration, ParseIntError> {
```
ecpullen: This should return an error type from this crate.

webern:
> Since the new function is only called with values that are previously checked with regex, running the function with a value that doesn't conform to the regex is undefined behavior. Should I "define" this behavior and add some unit tests for those cases, or only add tests for values that conform to the regex?

It's not super important in this case, so not worth losing a bunch of time on, but... 

It probably wouldn't hurt to use `?` instead of `unwrap_or_else([default])`. If you use anyhow::Error (via `crate::Error`) instead of ParseIntError then I think it's pretty easy to define the behavior of this function such that it errors on unexpected input. And it's always good to have unit tests that cover these "bad input" cases.

controller/src/constants.rs
ecpullen: Have you verified that d,h, and m are the only possible units for this data format?

controller/Cargo.toml
webern: This is why the builds are failing. The Cargo.lock needs to be updated and checked in to the commit.

mjsterckx: Yeah, I had that ready but I didn't want to do a bunch of force-pushes. It'll be in the next push.

controller/src/constants.rs
mjsterckx: d, h, m, and s are the only characters that can be in the string as per the validation regex. Since s (if present) is guaranteed to be the last character and the associated value would be 1, I am not including that in the vector.

mjsterckx:
^ Added unit tests and removed `parse_duration` and its dependencies from Cargo.lock

webern (approved):
I don't quite follow the parsing logic, but the tests convince me that it works.

stmcginnis (approved):
Looks like unit tests cover each case, including invalid cases. Might be nice to have the error message contain something more detailed as to why the parsing failed, but hopefully that should be pretty obvious to future us.

etungsten (commented):
Not sure if the parsing function belongs in the `constants` module. Maybe a new `utils.rs` is more appropriate.

ecpullen (approved):

PR #626, opened 2022-10-26
model: add constant for TestSys version

### LINKED ISSUE #542: Align released agent images and TestSys version
The system requires that the agents use the same TestSys version as the client. There should be a way to programmatically ensure this happens.

Somewhere in model, the version of testsys should be available so users can determine the testsys version that is being used. This will be useful for determining the location of the released testing images.

mjsterckx:
**Issue number:**

Closes #542 

**Description of changes:**

Added a public constant to `model/src/constants.rs` that stores the TestSys version. The constant is populated at compile-time and available at runtime.

**Testing done:**

Printed the constant from TestManager: output was `0.0.2`

**Terms of contribution:**

### DIFF
--- a/model/src/constants.rs
+++ b/model/src/constants.rs
@@ -14,6 +14,7 @@ macro_rules! testsys {
 pub const API_VERSION: &str = testsys!("v1");
 pub const NAMESPACE: &str = "testsys-bottlerocket-aws";
 pub const TESTSYS: &str = testsys!();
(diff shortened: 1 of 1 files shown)


webern (approved):

webern:
If this closes #542, @ecpullen do we have a related issue in the bottlerocket repo where we will use this to ensure version alignment?

ecpullen (approved):

PR #639, opened 2022-11-01
tools: fix Dockerfile build

jpculp:
**Description of changes:**

- Make sure that `clarify.toml` is in the expected root location and add some missing clarifications.

- Patch missing tag of **eksctl-anywhere** dependency.

**Testing done:**

Ran `make tools`.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #644, opened 2022-11-04
tools: clean up Dockerfile

jpculp:
**Description of changes:**

Re-organizes the Dockerfile to follow some best practices as well as splits up `eksctl-anywhere` into its own builder.

Also deletes the unused `build-rust` block and bumps the syntax version to **1.4.3**.

**Testing done:**

Built locally and tested with actions.

**Terms of contribution:**


tools/Dockerfile
```diff
@@ -81,25 +58,36 @@ RUN cp -p LICENSE /usr/share/licenses/eksctl && \
...
       /hashes/eksctl | sha512sum --check - && \
-    tar -xf eksctl_anywhere_${EKSA_VERSION}.tar.gz --strip-components 1 -C eksctl-anywhere && \
-    rm eksctl_anywhere_${EKSA_VERSION}.tar.gz && \
-    sed -i 's/v0.2.0-alpha.6/v0.2.1/' eksctl-anywhere/go.mod && \
-    rm eksctl-anywhere/go.sum
+    tar -xf "eksctl_anywhere_${EKSA_VERSION}.tar.gz" --strip-components 1 -C eksctl-anywhere && \
+    rm "eksctl_anywhere_${EKSA_VERSION}.tar.gz"
 
-USER root
+USER builder
```
etungsten, line 93: I'm pretty sure this has to be root or we'll run into permission issues.

jpculp: `/usr/share/licenses/eksctl-anywhere` is `chown`ed to **builder** so that shouldn't be the case. That said, it looks like this line needs to move to L81 which would explain any permissions issues you might have seen in the past (see L119). I'll post a revision!

jpculp:
- Move `USER builder` up a block.
- Add missing tag patch from #639.

jpculp:
I'll remove the missing tag patch and rebase after #645 merges.

jpculp:
Rebased on #645. Trying out `TESTSYS_BUILD_GOPROXY=https://proxy.golang.org/ make tools` to test, but I'm discovering a minor issue or two. I'll keep poking at it.

webern (approved):
LGTM - thank you for doing this.

@etungsten @mjsterckx @ecpullen any issues with this cleanup?

Looks like it needs some attention to get CI to pass, but if so, then looks good. (Maybe it just needs a rebase to pick up the build fix)

mjsterckx (approved):
LGTM (aside from the CI problem)

tools/Dockerfile
```diff
@@ -47,24 +32,16 @@ RUN mkdir -p /usr/share/licenses/eksctl && \
...
-
 USER builder
 WORKDIR /home/builder/
-# eksctl
-RUN mkdir eksctl && curl -L ${EKSCTL_SOURCE_URL} \
-      -o eksctl_${EKSCTL_VERSION}.tar.gz && \
-    grep eksctl_${EKSCTL_VERSION}.tar.gz \
+RUN mkdir eksctl && curl -L "${EKSCTL_SOURCE_URL}" \
+      -o "eksctl_${EKSCTL_VERSION}.tar.gz" && \
+    grep "eksctl_${EKSCTL_VERSION}.tar.gz" \
```
stmcginnis, line 40: Looks like there might be a mix of tabs and spaces here. The indentation probably looks the same in the local editor, but the diff view shows them at different levels.

Not related to your change, but if you do an update it would be nice to fix that to be consistent.

jpculp: No tabs here! This is by design to signify the continuation of a command. It might be more obvious if it looked like this instead:
```
RUN mkdir eksctl && \
    curl -L "${EKSCTL_SOURCE_URL}" \
      -o "eksctl_${EKSCTL_VERSION}.tar.gz" && \
    grep "eksctl_${EKSCTL_VERSION}.tar.gz" \
...
```

jpculp:
The permissions issues surrounding `eksctl-anywhere` stemmed from extracting the tarball to a shared directory that the builder user was not an owner of (`/tmp`).

(Also cleaned up some spacing to look a bit more intuitive.)

jpculp:
Rebased to get the updated `eksctl` / `eksctl-anywhere` and resolve conflicts.

tools/Dockerfile
jpculp: The issue was extracting the binary to `/tmp`:

```
#37 1.217 eksctl_anywhere_0.12.1_23_linux_amd64.tar.gz: OK
#37 1.686 tar: .: Cannot utime: Operation not permitted
#37 1.686 tar: .: Cannot change mode to rwxr-xr-t: Operation not permitted
```

jpculp:
Fixed character drop from the rebase.

tools/Dockerfile
```diff
@@ -81,23 +58,39 @@ RUN cp -p LICENSE /usr/share/licenses/eksctl && \
...
+    curl -L "${EKSA_SOURCE_URL}" \
+      -o "eksctl_anywhere_${EKSA_VERSION}.tar.gz" && \
+    grep "eksctl_anywhere_${EKSA_VERSION}.tar.gz" \
+      /hashes/eksctl-anywhere | sha512sum --check - && \
+    tar -xf "eksctl_anywhere_${EKSA_VERSION}.tar.gz" \
+      --strip-components 1 -C eksctl-anywhere && \
+    rm "eksctl_anywhere_${EKSA_VERSION}.tar.gz"
 
-USER root
+USER builder
```
stmcginnis, line 93: Set to builder already above. Doesn't look like this `USER` line is necessary now that line 82 was added.

tools/Dockerfile
```diff
@@ -1,70 +1,47 @@
...
+# =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
+# Shared build stage used to build Go binaries.
+FROM base as build-go
 
 ARG GOARCH
 ARG GOOS=linux
 ARG GOROOT="/usr/libexec/go"
 ARG GOPROXY
 
+USER builder
```
stmcginnis, line 19: Resets back to root on line 26, and no actions happen between here and then, so this looks unnecessary.

stmcginnis (approved):
Looks good. A couple minor things, but I don't consider them blocking.

etungsten (approved):
Thanks for the help here! Super appreciated!

tools/Dockerfile
jpculp: That should be adding the the proper **GOROOT** and **GOPROXY** to `builder`'s environment, which will get shared out to everything calling `build-go`, but it may be that the **ENV** would get shared regardless.

PR #646, opened 2022-11-04
agent: use Credentials instead of env variables

### LINKED ISSUE #583: agents: prevent role timeout by using SDKCredentials instead of `from_env`.
Currently in `agent/utils/src/aws.rs`, we assume the role by setting environment variables, instead we should at a minimum use `SDKCredentials` to create the config. Even better would be to create a credential provider for the `aws_config`.

Note: assume role can timeout and this can help prevent the timeout

mjsterckx:
**Issue number:**

Closes #583 

**Description of changes:**

Merged `aws_test_config` and `aws_resource_config` into one method `aws_config` and modified them to store and pass credentials using `Credentials` instead of environment variables. Added bool `setup_env` to both methods to determine whether or not the credentials should be set as environment variables (which should happen when, for example, running `sonobuoy` tests). Added `assume_role_session_duration` argument to `aws_config` which was previously only used in `aws_test_config`. Refactored `aws.rs` to reduce code duplication.

**Testing done:**

Ran an ECS test and an EKS test, with and without assumed role, all 4 of which passed.

**Terms of contribution:**

### DIFF
--- a/agent/utils/src/aws.rs
+++ b/agent/utils/src/aws.rs
@@ -1,35 +1,28 @@
-use crate::constants::DEFAULT_REGION;
+use crate::constants::{DEFAULT_ASSUME_ROLE_SESSION_DURATION, DEFAULT_REGION};
 use crate::error::{self, Error};
-use aws_config::meta::region::RegionProviderChain;
+use agent_common::secrets::SecretsReader;
+use aws_config::default_provider::credentials::default_provider;
+use aws_config::sts::AssumeRoleProvider;
 use aws_config::RetryConfig;
 use aws_sdk_sts::Region;
 use aws_smithy_types::retry::RetryMode;
+use aws_types::credentials::{Credentials, SharedCredentialsProvider};
 use aws_types::SdkConfig;
 use log::info;
 use model::SecretName;
-use resource_agent::clients::InfoClient;
-use resource_agent::provider::{ProviderResult, Resources};
 use snafu::{OptionExt, ResultExt};
 use std::env;
-use test_agent::Runner;
+use std::time::Duration;
 
 /// Set up the config for aws calls using `aws_secret_name` if provided and `sts::assume_role`
-/// if a role arn is provided.
-pub async fn aws_test_config<R>(
-    runner: &R,
-    aws_secret_name: &Option<SecretName>,
+/// if a role arn is provided. Set credentials as environment variables if `setup_env` is true
+pub async fn aws_config(
+    aws_secret_name: &Option<&SecretName>,
     assume_role: &Option<String>,
     assume_role_session_duration: &Option<i32>,
     region: &Option<String>,
-) -> Result<SdkConfig, Error>
-where
-    R: Runner,
-{
-    if let Some(aws_secret_name) = aws_secret_name {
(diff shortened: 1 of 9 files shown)


agent/utils/src/aws.rs
```diff
@@ -1,35 +1,28 @@
...
-) -> Result<SdkConfig, Error>
-where
-    R: Runner,
-{
-    if let Some(aws_secret_name) = aws_secret_name {
-        info!("Adding secret '{}' to the environment", aws_secret_name);
-        setup_test_env(runner, aws_secret_name).await?;
-    }
-
+    setup_env: bool,
```
ecpullen, line 24: Can you add a comment about what this bool does?

agent/utils/src/aws.rs
```diff
@@ -38,19 +31,49 @@ where
...
+    let mut config_loader = aws_config::from_env().retry_config(
+        RetryConfig::standard()
+            .with_retry_mode(RetryMode::Adaptive)
+            .with_max_attempts(15),
+    );
+    let base_provider = match aws_secret_name {
+        Some(aws_secret_name) => {
+            let (access_key_id, secret_access_key, session_token) =
+                get_secret_values(aws_secret_name)?;
+            if setup_env {
```
ecpullen: `if setup_env && assumre_role.is_none()`

mjsterckx:
^ Addressed ecpullen@ comments

agent/utils/src/aws.rs
```diff
@@ -64,49 +87,34 @@ where
...
+            ),
+            &String::from(
+                credentials
+                    .secret_access_key()
+                    .context(error::CredentialsMissingSnafu { role_arn })?,
+            ),
+            &Some(String::from(
+                credentials
+                    .session_token()
+                    .context(error::CredentialsMissingSnafu { role_arn })?,
```
ecpullen, line 104: ~~The session token should not be required.~~
Misread the code.

agent/utils/src/aws.rs
```diff
@@ -64,49 +87,34 @@ where
...
 }
 
-/// Set up AWS credential secrets in a runner's process's environment
-pub async fn setup_test_env<R>(runner: &R, aws_secret_name: &SecretName) -> Result<(), Error>
-where
-    R: Runner,
-{
-    let aws_secret = runner
+fn get_secret_values(
+    aws_secret_name: &&SecretName,
```
ecpullen: Why is this `&&SecretName`?

agent/utils/src/aws.rs
```diff
@@ -64,49 +87,34 @@ where
...
+) -> Result<(String, String, Option<String>), Error> {
+    let reader = SecretsReader::new();
+    let aws_secret = reader
         .get_secret(aws_secret_name)
-        .context(error::SecretMissingSnafu)?;
-
+        .context(error::SecretMissingSnafu {})?;
```
ecpullen: Why did the `{}` get added here?

ecpullen (changes requested):

mjsterckx:
^ Fixed typos

ecpullen (approved):
🚢

stmcginnis (approved):
Looks good to me. If not, we can fix it later. ;)

PR #669, opened 2022-11-23
Add k8s workload testing agent

### LINKED ISSUE #616: workload test agent for k8s/sonobuoy
Create the workload test agent, as described in #419, for k8s/sonobuoy testing.

stmcginnis:
**Issue number:**

Closes: #616 

**Description of changes:**

This adds a test agent for running cluster workload tests. These are tests that run some sort of workload on the cluster to verify system functionality.

**Testing done:**

- Created test container image, manually stepped through the sonobuoy commands in the agent and verified the job runs and is successful on aws-k8s-nvidia nodes.
- Followed the [developer instructions for running a test](https://github.com/bottlerocket-os/bottlerocket-test-system/blob/develop/docs/DEVELOPER.md#running-a-test), replacing `example-test-agent` with `k8s-workload-agent` and its settings. Unable to run the NVIDIA tests against a local kind cluster, but this verified things were deployed and status reported back. This along with the previous step that verified the workload test itself indicates things would perform correctly in an actual test (non-kind) cluster.
- More testing to come!

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -198,3 +198,34 @@ COPY --from=build-src /src/bottlerocket/agents/src/bin/migration-test-agent/ssm-
 COPY --from=build-src /usr/share/licenses/testsys /licenses/testsys
 
 ENTRYPOINT ["./migration-test-agent"]
+
+# =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
+# Builds the Kubernetes Workload test agent image
+FROM public.ecr.aws/amazonlinux/amazonlinux:2 AS k8s-workload-agent
+ARG ARCH
+
+# TODO remove unzip once aws-cli moves out
+RUN yum install -y unzip iproute && yum clean all
+ARG AWS_CLI_URL=https://awscli.amazonaws.com/awscli-exe-linux-${ARCH}.zip
+
+# Copy aws-iam-authenticator
+COPY --from=tools /aws-iam-authenticator /usr/bin/aws-iam-authenticator
+COPY --from=tools /licenses/aws-iam-authenticator /licenses/aws-iam-authenticator
+
+# TODO move this out, get hashes, and attribute licenses
+# Download aws-cli
+RUN temp_dir="$(mktemp -d --suffix aws-cli)" && \
+    curl -fsSL "${AWS_CLI_URL}" -o "${temp_dir}/awscliv2.zip" && \
+    unzip "${temp_dir}/awscliv2.zip" -d "${temp_dir}" && \
+    ${temp_dir}/aws/install && \
+    rm -rf ${temp_dir}
+
+# Copy sonobuoy
+COPY --from=tools /sonobuoy /usr/bin/sonobuoy
+COPY --from=tools /licenses/sonobuoy /licenses/sonobuoy
+
+# Copy k8s-workload-agent
+COPY --from=build-src /src/bottlerocket/agents/bin/k8s-workload-agent ./
+COPY --from=build-src /usr/share/licenses/testsys /licenses/testsys
+
+ENTRYPOINT ["./k8s-workload-agent"]
--- a/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
+++ b/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
@@ -1,0 +1,111 @@
+/*!
+
+This is a test-agent for running workload tests on Kubernetes.
+It expects to be run in a pod launched by the TestSys controller.
+
+You can configure the workload agent to run different types of plugins and tests.
+See `WorkloadConfig` for the different configuration values.
+
+To build the container for the workload test agent, run `make k8s-workload-agent-image` from the
+root directory of this repository.
+
+Here is an example manifest for deploying the test definition for the workload test agent to a K8s cluster:
+
+```yaml
+apiVersion: testsys.system/v1
+kind: Test
+metadata:
+  name: workload-full
+  namespace: testsys
+spec:
+  agent:
+    configuration:
+      kubeconfigBase64: <Base64 encoded kubeconfig for the test cluster workload runs the tests in>
+      plugins:
+      - name: nvidia-workload
+        image: testsys-nvidia-workload-test:v0.0.3
+    image: <your k8s-workload-agent image URI>
+    name: workload-test-agent
+    keepRunning: true
+  resources: {}
+```
+
+!*/
+
+use agent_utils::{base64_decode_write_file, init_agent_logger};
+use async_trait::async_trait;
+use bottlerocket_agents::constants::TEST_CLUSTER_KUBECONFIG_PATH;
+use bottlerocket_agents::error::Error;
+use bottlerocket_agents::workload::{delete_workload, rerun_failed_workload, run_workload};
+use bottlerocket_types::agent_config::WorkloadConfig;
+use log::info;
+use model::TestResults;
+use std::path::PathBuf;
+use test_agent::{BootstrapData, ClientError, DefaultClient, Spec, TestAgent};
+
+struct WorkloadTestRunner {
+    config: WorkloadConfig,
+    results_dir: PathBuf,
+}
+
+#[async_trait]
+impl test_agent::Runner for WorkloadTestRunner {
+    type C = WorkloadConfig;
+    type E = Error;
+
+    async fn new(spec: Spec<Self::C>) -> Result<Self, Self::E> {
+        info!("Initializing Workload test agent...");
+        Ok(Self {
+            config: spec.configuration,
+            results_dir: spec.results_dir,
+        })
+    }
+
+    async fn run(&mut self) -> Result<TestResults, Self::E> {
+        info!("Decoding kubeconfig for test cluster");
+        base64_decode_write_file(&self.config.kubeconfig_base64, TEST_CLUSTER_KUBECONFIG_PATH)
+            .await?;
+        info!("Stored kubeconfig in {}", TEST_CLUSTER_KUBECONFIG_PATH);
+
+        run_workload(
+            TEST_CLUSTER_KUBECONFIG_PATH,
+            &self.config,
+            &self.results_dir,
+        )
+        .await
+    }
+
+    async fn rerun_failed(&mut self, _prev_results: &TestResults) -> Result<TestResults, Self::E> {
+        delete_workload(TEST_CLUSTER_KUBECONFIG_PATH).await?;
+
+        info!("Decoding kubeconfig for test cluster");
+        base64_decode_write_file(&self.config.kubeconfig_base64, TEST_CLUSTER_KUBECONFIG_PATH)
+            .await?;
+        info!("Stored kubeconfig in {}", TEST_CLUSTER_KUBECONFIG_PATH);
+
+        rerun_failed_workload(TEST_CLUSTER_KUBECONFIG_PATH, &self.results_dir).await
+    }
+
+    async fn terminate(&mut self) -> Result<(), Self::E> {
+        delete_workload(TEST_CLUSTER_KUBECONFIG_PATH).await
+    }
+}
+
+#[tokio::main]
+async fn main() {
+    init_agent_logger(env!("CARGO_CRATE_NAME"), None);
+    if let Err(e) = run().await {
+        eprintln!("{}", e);
+        std::process::exit(1);
+    }
+}
+
+async fn run() -> Result<(), test_agent::error::Error<ClientError, Error>> {
+    let mut agent = TestAgent::<DefaultClient, WorkloadTestRunner>::new(
+        BootstrapData::from_env().unwrap_or_else(|_| BootstrapData {
+            test_name: "workload_test".to_string(),
+        }),
+    )
+    .await?;
+    agent.run().await
+}
--- a/bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
+++ b/bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
@@ -127,7 +127,6 @@ impl test_agent::Runner for SonobuoyTestRunner {
         rerun_failed_sonobuoy(
             TEST_CLUSTER_KUBECONFIG_PATH,
             e2e_repo_config,
-            &self.config,
             &self.results_dir,
         )
         .await
--- a/bottlerocket/agents/src/sonobuoy.rs
+++ b/bottlerocket/agents/src/sonobuoy.rs
@@ -2,7 +2,9 @@ use crate::error;
 use bottlerocket_types::agent_config::{SonobuoyConfig, SONOBUOY_RESULTS_FILENAME};
 use log::{error, info, trace};
 use model::{Outcome, TestResults};
+use serde_json::json;
 use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::HashMap;
 use std::path::Path;
 use std::process::Command;
 
@@ -56,14 +58,13 @@ pub async fn run_sonobuoy(
     // TODO - log something or check what happened?
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
-    results_sonobuoy(kubeconfig_path, sonobuoy_config, results_dir)
+    results_sonobuoy(kubeconfig_path, results_dir)
 }
 
 /// Reruns the the failed tests from sonobuoy conformance that has already run in this agent.
 pub async fn rerun_failed_sonobuoy(
     kubeconfig_path: &str,
     e2e_repo_config_path: Option<&str>,
-    sonobuoy_config: &SonobuoyConfig,
     results_dir: &Path,
 ) -> Result<TestResults, error::Error> {
     let kubeconfig_arg = vec!["--kubeconfig", kubeconfig_path];
@@ -91,13 +92,12 @@ pub async fn rerun_failed_sonobuoy(
     // TODO - log something or check what happened?
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
-    results_sonobuoy(kubeconfig_path, sonobuoy_config, results_dir)
+    results_sonobuoy(kubeconfig_path, results_dir)
 }
 
 /// Retrieve the results from a sonobuoy test and convert them into `TestResults`.
 pub fn results_sonobuoy(
     kubeconfig_path: &str,
-    sonobuoy_config: &SonobuoyConfig,
     results_dir: &Path,
 ) -> Result<TestResults, error::Error> {
     let kubeconfig_arg = vec!["--kubeconfig", kubeconfig_path];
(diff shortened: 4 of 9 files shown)


Dockerfile
```diff
@@ -198,3 +198,34 @@ COPY --from=build-src /src/bottlerocket/agents/src/bin/migration-test-agent/ssm-
...
+RUN yum install -y unzip iproute && yum clean all
+ARG AWS_CLI_URL=https://awscli.amazonaws.com/awscli-exe-linux-${ARCH}.zip
+
+# Copy aws-iam-authenticator
+COPY --from=tools /aws-iam-authenticator /usr/bin/aws-iam-authenticator
+COPY --from=tools /licenses/aws-iam-authenticator /licenses/aws-iam-authenticator
+
+# TODO move this out, get hashes, and attribute licenses
+# Download aws-cli
+RUN temp_dir="$(mktemp -d --suffix aws-cli)" && \
```
webern, line 217: Do you actually need the aws cli? This is the only place I see it.

webern (commented):
Cool

Dockerfile
stmcginnis: I'm not actually sure if this is used internally some how. Now that I've gotten further than when I added this, I'm guessing no. I copied this portion from the sonobuoy definition since these workload tests are basically the same thing, just executed differently. If it is not needed here, then it would not be needed their either and we should clean it up.

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
```diff
@@ -0,0 +1,116 @@
...
+  name: workload-full
+  namespace: testsys
+spec:
+  agent:
+    configuration:
+      kubeconfigBase64: <Base64 encoded kubeconfig for the test cluster workload runs the tests in>
+      plugins:
+      - name: nvidia-workload
+        image: testsys-nvidia-workload-test:v0.0.3
+    image: <your k8s-workload-agent image URI>
```
ecpullen, line 27: Should we put the expected image URI?

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -97,7 +98,7 @@ pub async fn rerun_failed_sonobuoy(
 /// Retrieve the results from a sonobuoy test and convert them into `TestResults`.
 pub fn results_sonobuoy(
     kubeconfig_path: &str,
-    sonobuoy_config: &SonobuoyConfig,
+    _: &SonobuoyConfig,
```
ecpullen: I think we can get rid of this arg now.

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+        .args(kubeconfig_arg.to_owned())
+        .arg("run")
+        .arg("--wait")
+        .arg("--namespace")
+        .arg("testsys-workload")
+        .args(plugin_test_args)
+        .status()
+        .context(error::WorkloadProcessSnafu)?;
+    info!("Workload testing has completed, checking results");
+
```
ecpullen: It probably shouldn't be in this pr, but it would be nice to not use `--wait` on the run command incase connection to the cluster is broken for a short period of time. It would be better to `sonobuoy run` and then routinely check `sonobuoy status`. (This also should be fixed in the sonobuoy agent)

bottlerocket/types/src/agent_config.rs
```diff
@@ -12,6 +12,7 @@ use std::str::FromStr;
 pub const AWS_CREDENTIALS_SECRET_NAME: &str = "awsCredentials";
 pub const SONOBUOY_RESULTS_FILENAME: &str = "sonobuoy-results.tar.gz";
 pub const VSPHERE_CREDENTIALS_SECRET_NAME: &str = "vsphereCredentials";
+pub const WORKLOAD_RESULTS_FILENAME: &str = "workload-results.tar.gz";
```
ecpullen: Can this not just be `SONOBUOY_RESULTS_FILENAME`?

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
```diff
@@ -0,0 +1,116 @@
...
+    async fn new(spec: Spec<Self::C>) -> Result<Self, Self::E> {
+        info!("Initializing Workload test agent...");
+        Ok(Self {
+            config: spec.configuration,
+            results_dir: spec.results_dir,
+        })
+    }
+
+    async fn run(&mut self) -> Result<TestResults, Self::E> {
+        debug!("Decoding kubeconfig for test cluster");
```
webern: I think we should be more verbose with our logging since it currently isn't easy to change the log level for these containers. And when something goes wrong these provide nice anchor points in the code.

```suggestion
        info!("Decoding kubeconfig for test cluster");
```

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -143,72 +144,102 @@ pub fn results_sonobuoy(
...
+/// process_sonobuoy_test_results parses the output from `sonobuoy status --json` output and gets
+/// the overall status of the plugin results.
+pub(crate) fn process_sonobuoy_test_results(
+    run_status: &serde_json::Value,
+) -> Result<TestResults, error::Error> {
+    let mut num_passed: u64 = 0;
+    let mut num_failed: u64 = 0;
+    let mut num_skipped: u64 = 0;
+    let mut progress = Vec::new();
+    let mut outcome_summary = HashMap::from([
```
webern, line 158: I don't love this map. A struct that implements default would be nicer imo.

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -143,72 +144,102 @@ pub fn results_sonobuoy(
...
+        ("pass", 0),
+        ("passed", 0),
+        ("fail", 0),
+        ("failed", 0),
+        ("timeout", 0),
+        ("timed-out", 0),
```
webern, line 164: In the theoretical struct you wouldn't need separate fields for the alternate spellings.

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -143,72 +144,102 @@ pub fn results_sonobuoy(
...
-        .unwrap_or(0);
-
-    let num_skipped = result_counts
-        .get("skipped")
-        .map(|v| v.as_u64().unwrap_or(0))
-        .unwrap_or(0);
+        .context(error::MissingSonobuoyStatusFieldSnafu { field: "plugins" })?;
+
+    for result in plugin_results {
+        let plugin = result
```
webern, line 175: Do we have somewhere in `TestResults` that we could report the list of plugins that reported failuers? I guess we could put in OtherInfo along with the progress report(s)?

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -227,3 +258,138 @@ pub async fn delete_sonobuoy(kubeconfig_path: &str) -> Result<(), error::Error>
 
     Ok(())
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+#[cfg(test)]
+mod test_sonobuoy {
+    use super::*;
+    // use serde::Serialize;
```
webern:
```suggestion
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+            .context(error::WorkloadProcessSnafu)?;
+        ensure!(
+            output.status.success(),
+            error::WorkloadPluginSnafu {
+                plugin: plugin.name.clone()
+            }
+        );
+
+        // Write out the output to a file we can reference later
+        let mut f = File::create(plugin_yaml.clone()).context(error::WorkloadProcessSnafu)?;
```
webern: Seeing this path could save somebody some troubleshooting time someday.

```suggestion        
        let plugin_yaml =  PathBuf::from(".")
             .join(format!("{}-plugin.yaml", plugin.name))
             .canonicalize()
             .context(error::BadPathSnafu{ path: plugin_yaml.clone())?;
        let mut f = File::create(&plugin_yaml).context(error::FileWrite{ path: plugin_yaml.clone()})?;
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+pub async fn run_workload(
+    kubeconfig_path: &str,
+    workload_config: &WorkloadConfig,
+    results_dir: &Path,
+) -> Result<TestResults, error::Error> {
+    info!("Processing workload test plugins");
+    let mut plugin_test_args: Vec<String> = Vec::new();
+    for (id, plugin) in workload_config.plugins.iter().enumerate() {
+        info!("Initializing test {}-{}", id, plugin.name);
+        let plugin_yaml = format!("{}-plugin.yaml", plugin.name);
```
webern: I would put this at its usage site.

```suggestion
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        );
+
+        // Write out the output to a file we can reference later
+        let mut f = File::create(plugin_yaml.clone()).context(error::WorkloadProcessSnafu)?;
+        f.write_all(&output.stdout)
+            .context(error::WorkloadProcessSnafu)?;
+
+        // Add plugin to the arguments to be passed to sonobuoy run
+        plugin_test_args.append(&mut vec!["--plugin".to_string(), plugin_yaml.to_string()]);
```
webern: Because I suggested making it a path...

```suggestion
        plugin_test_args.append(&mut vec!["--plugin".to_string(), plugin_yaml.display().to_string()]);
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+        .arg("--wait")
+        .arg("--namespace")
+        .arg("testsys-workload")
+        .args(plugin_test_args)
+        .status()
+        .context(error::WorkloadProcessSnafu)?;
+    info!("Workload testing has completed, checking results");
+
+    // TODO - log something or check what happened?
+    ensure!(status.success(), error::WorkloadRunSnafu);
```
webern: There are a couple of options here. You can either stream to stderr and stdout while the command is running, or you can call `output()` instead of `status()` to get the stdout/stderr. The latter would look something like this:

```suggestion
    let output = Command::new(SONOBUOY_BIN_PATH)
        .args(kubeconfig_arg.to_owned())
        .arg("run")
        .arg("--wait")
        .arg("--namespace")
        .arg("testsys-workload")
        .args(plugin_test_args)
        .output()
        .context(error::WorkloadProcessSnafu)?;
    info!("Workload testing has completed, checking results");

    ensure!(
        output.status.success(),
        error::WorkloadRunSnafu{
            exit_code: output.code().unwrap_or(1),
            stdout: output.stdout.convert_to_utf8(), // can't remember how
            stderr: output.stderr.convert_to_utf8(), // can't remember how
        }
    );
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,155 @@
...
+        .arg("--wait")
+        .arg("--namespace")
+        .arg("testsys-workload")
+        .arg("--rerun-failed")
+        .arg(results_filepath.as_os_str())
+        .status()
+        .context(error::WorkloadProcessSnafu)?;
+    info!("Workload testing has completed, checking results");
+
+    // TODO - log something or check what happened?
```
webern: See suggestion at the other TODO site.

Dockerfile
stmcginnis: @etungsten it looks like [you added this for the sonobuoy image](https://github.com/bottlerocket-os/bottlerocket-test-system/commit/f0f4604e2a201fdb12ea13f3088404c94cdfd29d). Do you know, or remember, if this is actually needed for something when running sonobuoy?

etungsten: EKS clusters kubeconfigs uses either `aws-cli` or `aws-iam-autenticator` to authenticate to the cluster and it expect those tools to be in path. See https://docs.aws.amazon.com/eks/latest/userguide/create-kubeconfig.html. When passing `--kubeconfig` to sonobuoy, if that kubeconfig is using `aws-cli` as the exec plugin, then sonobuoy needs it to retrieve the client token for talking to the cluster.

stmcginnis: Awesome, thanks for the details Erikson. So since the workload agent testing is based on sonobuoy plugin usage I believe it needs the same things the sonobuoy agent does.

etungsten: That being said, now that there are two container images using aws-cli, we should move this to the `tools` image and copy the binary in here like we do for `aws-iam-authenticator` above.

stmcginnis: Add https://github.com/bottlerocket-os/bottlerocket-test-system/issues/675 to track moving aws cli to `tools` to keep that work separate from this PR.

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
stmcginnis: We don't include that in any of the other test agent examples. Maybe after this we can go through and update them all of we have a concrete image URI for all agents.

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
stmcginnis: Updated.

bottlerocket/agents/src/sonobuoy.rs
stmcginnis: Done.

bottlerocket/types/src/agent_config.rs
stmcginnis: Done.

bottlerocket/agents/src/sonobuoy.rs
stmcginnis: Done.

bottlerocket/agents/src/workload.rs
stmcginnis: Done, with some slight modifications to make this work right. Please confirm though.

bottlerocket/agents/src/workload.rs
stmcginnis: Done

bottlerocket/agents/src/workload.rs
stmcginnis: Done.

bottlerocket/agents/src/sonobuoy.rs
stmcginnis: So maybe something like:

```rust
if (results_status == "fail" || results_status == "failed") && progress_status.is_empty() {
    progress.push(format!("{}: Failed", plugin);
}
```

bottlerocket/agents/src/sonobuoy.rs
stmcginnis: I may need a clearer picture of what you're envisioning here. It looks like plugins can report either variation for each result, so somewhere we need to evaluate them.

My thought with this structure was we wouldn't have to check each value every time, just collect it all up and get the totals at the end.

bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
```diff
@@ -127,7 +127,6 @@ impl test_agent::Runner for SonobuoyTestRunner {
         rerun_failed_sonobuoy(
             TEST_CLUSTER_KUBECONFIG_PATH,
             e2e_repo_config,
-            &self.config,
```
ecpullen, line 130: Did you verify that the original sonobuoy test agent still worked with this change?

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -143,72 +142,102 @@ pub fn results_sonobuoy(
...
-        .get("skipped")
-        .map(|v| v.as_u64().unwrap_or(0))
-        .unwrap_or(0);
+        .context(error::MissingSonobuoyStatusFieldSnafu { field: "plugins" })?;
+
+    for result in plugin_results {
+        let plugin = result
+            .get("plugin")
+            .map(|value| value.as_str().unwrap())
+            .unwrap_or_else(|| "");
```
ecpullen: Usually we try to avoid `unwrap`s
```suggestion
            .and_then(|value| value.as_str())
            .unwrap_or_default();
```

bottlerocket/agents/src/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    info!("Running workload");
+    let kubeconfig_arg = vec!["--kubeconfig", kubeconfig_path];
+    let output = Command::new(SONOBUOY_BIN_PATH)
+        .args(kubeconfig_arg.to_owned())
+        .arg("run")
+        .arg("--wait")
+        .arg("--namespace")
+        .arg("testsys-workload")
```
ecpullen, line 67: Does this mean we could run a normal `sonobuoy` test and `workload` test on the same cluster at the same time? (Since they would be in different namespaces)

ecpullen (approved):
One small nit

bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
stmcginnis: I will double check that and report back.

bottlerocket/agents/src/workload.rs
stmcginnis: Yep, that's the goal. Since this will be placed into a different non-default namespace from sonobuoy, it should be able to be run at the same time without interfering with each other. After double checking based on the comment above, I will also double check this and report back too.

etungsten (approved):

bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
stmcginnis: Ran tests with both sonobuoy and the workload tests. Both worked fine together. Controllers are both in the `testsys` namespace, but test jobs themselves are created in the separate namespaces with the workload jobs running in the `testsys-workload` namespace.

bottlerocket/agents/src/workload.rs
stmcginnis: Ran additional tests and verified, see comment above in `sonobuoy.rs`.

bottlerocket/agents/src/sonobuoy.rs
stmcginnis: I realized we had a better pattern already established with the `result_status` field below. I've updated this and `progress_status` to keep consistent with that.

webern (changes requested):
Looks like the "integration" tests are not happy.

webern (approved):
Was the failed integ test run just a fluke?

stmcginnis:
> Was the failed integ test run just a fluke?

No, it was a legitimate unit test failure due to a quick "safe" change I had made in the last update. Sorry, I thought I had commented on this PR, but no comment here. (Where did I post that?!). If you look at the Compare from the last force-push, it was a small change due to an optional value we look for in the results.

PR #683, opened 2022-12-02
ec2-agent/vsphere-vm-agent: allow custom userdata

### LINKED ISSUE #627: ec2-agent: Allow custom userdata when launching instances
Allow the user to provide custom user-data TOML that will be appended/merged into the user-data settings for test nodes created by the ec2 resource agent.

mjsterckx:
**Issue number:**

Closes #627 

**Description of changes:**

The `Ec2Config` and `VSphereVmConfig` are modified to allow an optional `CustomUserData` field. This field contains either a `Merge` or a `Replace` value with encoded userdata. The userdata either replaces the default userdata (if the `Replace` mode is used) or is parsed into `toml::Value` and merged with the default userdata (if the `Merge` mode is used).

**Testing done:**

- Unit tests for `merge_values`
- Ran an ECS test with this TOML file as custom userdata (in merge mode):
```
[settings.ecs]
loglevel = "warn"

[settings]
motd = "Custom userdata"
```

Test passed and instance userdata looks like this:
```
[settings]
motd = "Custom userdata"

[settings.ecs]
cluster = "external-cluster"
loglevel = "warn"
```
- Ran an EKS test with this TOML file as custom userdata (in merge mode):
```
[settings.kubernetes]
log-level = 3

[settings]
motd = "Custom userdata"
```

Test passed and instance userdata looks like this:
```
[settings]
motd = "Custom userdata"

[settings.kubernetes]
api-server = "https://ECEE40E694F1EA9026E6415225549739.gr7.us-east-1.eks.amazonaws.com"
cluster-certificate = "****************"
cluster-dns-ip = "10.100.0.10"
cluster-name = "eks-test-cluster"
log-level = 3

[settings.updates]
ignore-waves = true
```
- Ran a VMware test with this TOML file as custom userdata (in merge mode):
```
[settings.updates]
ignore-waves = true

[settings]
motd = "Custom userdata"
```

Test passed and userdata looks like this:
```
[settings]
motd = "Custom userdata"
[settings.host-containers.control]
enabled = true
user-data = "eyJzc20iOnsiYWN0aXZhdGlvbi1pZCI6IjY5YmY5YzgzLWEwYTktNDA5My04M2M2LTQxM2FhZTRiNDI1ZiIsImFjdGl2YXRpb24tY29kZSI6IlpyOGJ4ZVA5Q1h6SzJSZ0ZsZXlsIiwicmVnaW9uIjoidXMtd2VzdC0yIn19"

[settings.kubernetes]
api-server = "https://***.**.**.***:****/"
bootstrap-token = "****************"
cluster-certificate = "****************"
cluster-dns-ip = "**.**.*.**"

[settings.updates]
ignore-waves = true
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
+++ b/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
@@ -8,7 +8,10 @@ use aws_sdk_ec2::model::{
 };
 use aws_sdk_ec2::output::RunInstancesOutput;
 use aws_sdk_ec2::types::SdkError;
-use bottlerocket_types::agent_config::{ClusterType, Ec2Config, AWS_CREDENTIALS_SECRET_NAME};
+use bottlerocket_agents::userdata::{decode_to_string, merge_values};
+use bottlerocket_types::agent_config::{
+    ClusterType, CustomUserData, Ec2Config, AWS_CREDENTIALS_SECRET_NAME,
+};
 use log::{debug, info, trace, warn};
 use model::{Configuration, SecretName};
 use resource_agent::clients::InfoClient;
@@ -20,6 +23,7 @@ use std::collections::HashSet;
 use std::fmt::Debug;
 use std::iter::FromIterator;
 use std::time::Duration;
+use toml::Value;
 use uuid::Uuid;
 
 /// The default number of instances to spin up.
@@ -205,6 +209,7 @@ async fn run_instances(
                     &spec.configuration.endpoint,
                     &spec.configuration.certificate,
                     &spec.configuration.cluster_dns_ip,
+                    &spec.configuration.custom_user_data,
                     memo,
                 )?)
                 .iam_instance_profile(
@@ -335,35 +340,76 @@ fn userdata(
     endpoint: &Option<String>,
     certificate: &Option<String>,
     cluster_dns_ip: &Option<String>,
+    custom_user_data: &Option<CustomUserData>,
     memo: &ProductionMemo,
 ) -> ProviderResult<String> {
-    Ok(match cluster_type {
-        ClusterType::Eks => base64::encode(format!(
-            r#"[settings.updates]
+    let default_userdata = match cluster_type {
+        ClusterType::Eks => {
+            default_eks_userdata(cluster_name, endpoint, certificate, cluster_dns_ip, memo)?
+        }
+        ClusterType::Ecs => default_ecs_userdata(cluster_name),
+    };
+
+    let custom_userdata = if let Some(value) = custom_user_data {
+        value
+    } else {
+        return Ok(default_userdata);
+    };
+
+    match custom_userdata {
+        CustomUserData::Replace { encoded_userdata } => Ok(encoded_userdata.to_string()),
+        CustomUserData::Merge { encoded_userdata } => {
+            let merge_into = &mut decode_to_string(&default_userdata)?
+                .parse::<Value>()
+                .context(Resources::Clear, "Failed to parse TOML")?;
+            let merge_from = decode_to_string(encoded_userdata)?
+                .parse::<Value>()
+                .context(Resources::Clear, "Failed to parse TOML")?;
+            merge_values(&merge_from, merge_into)
+                .context(Resources::Clear, "Failed to merge TOML")?;
+            Ok(base64::encode(toml::to_string(merge_into).context(
+                Resources::Clear,
+                "Failed to serialize merged TOML",
+            )?))
+        }
+    }
+}
+
+fn default_eks_userdata(
+    cluster_name: &str,
+    endpoint: &Option<String>,
+    certificate: &Option<String>,
+    cluster_dns_ip: &Option<String>,
+    memo: &ProductionMemo,
+) -> Result<String, ProviderError> {
+    Ok(base64::encode(format!(
+        r#"[settings.updates]
 ignore-waves = true
-    
+
 [settings.kubernetes]
 api-server = "{}"
 cluster-name = "{}"
 cluster-certificate = "{}"
 cluster-dns-ip = "{}""#,
-            endpoint
-                .as_ref()
-                .context(memo, "Server endpoint is required for eks clusters.")?,
-            cluster_name,
-            certificate
-                .as_ref()
-                .context(memo, "Cluster certificate is required for eks clusters.")?,
-            cluster_dns_ip
-                .as_ref()
-                .context(memo, "Cluster DNS IP is required for eks clusters.")?,
-        )),
-        ClusterType::Ecs => base64::encode(format!(
-            r#"[settings.ecs]
+        endpoint
+            .as_ref()
+            .context(memo, "Server endpoint is required for eks clusters.")?,
+        cluster_name,
+        certificate
+            .as_ref()
+            .context(memo, "Cluster certificate is required for eks clusters.")?,
+        cluster_dns_ip
+            .as_ref()
+            .context(memo, "Cluster DNS IP is required for eks clusters.")?,
+    )))
+}
+
+fn default_ecs_userdata(cluster_name: &str) -> String {
+    base64::encode(format!(
+        r#"[settings.ecs]
 cluster = "{}""#,
-            cluster_name,
-        )),
-    })
+        cluster_name,
+    ))
 }
 
 #[derive(Debug)]
--- a/bottlerocket/agents/src/bin/vsphere-vm-resource-agent/vsphere_vm_provider.rs
+++ b/bottlerocket/agents/src/bin/vsphere-vm-resource-agent/vsphere_vm_provider.rs
@@ -3,9 +3,10 @@ use agent_utils::aws::aws_config;
 use agent_utils::base64_decode_write_file;
 use bottlerocket_agents::constants::TEST_CLUSTER_KUBECONFIG_PATH;
 use bottlerocket_agents::tuf::{download_target, tuf_repo_urls};
+use bottlerocket_agents::userdata::{decode_to_string, merge_values};
 use bottlerocket_agents::vsphere::vsphere_credentials;
 use bottlerocket_types::agent_config::{
-    VSphereVmConfig, AWS_CREDENTIALS_SECRET_NAME, VSPHERE_CREDENTIALS_SECRET_NAME,
+    CustomUserData, VSphereVmConfig, AWS_CREDENTIALS_SECRET_NAME, VSPHERE_CREDENTIALS_SECRET_NAME,
 };
 use k8s_openapi::api::core::v1::Service;
 use kube::api::ListParams;
@@ -27,6 +28,7 @@ use std::fs::File;
 use std::path::Path;
 use std::process::Command;
 use std::time::Duration;
+use toml::Value;
 
 /// The default number of VMs to spin up.
 const DEFAULT_VM_COUNT: i32 = 2;
@@ -268,6 +270,8 @@ impl Create for VMCreator {
         let encoded_control_host_ctr_userdata =
             base64::encode(control_host_ctr_userdata.to_string());
 
+        let custom_user_data = spec.configuration.custom_user_data;
+
         // Base64 encode userdata
         let userdata = userdata(
             &vsphere_cluster.control_plane_endpoint_ip,
@@ -275,7 +279,8 @@ impl Create for VMCreator {
             bootstrap_token,
             cluster_certificate,
             &encoded_control_host_ctr_userdata,
-        );
+            &custom_user_data,
+        )?;
 
         info!("Launching {} Bottlerocket worker nodes", vm_count);
         for i in 0..vm_count {
@@ -403,6 +408,47 @@ impl Create for VMCreator {
 }
 
 fn userdata(
+    endpoint: &str,
+    cluster_dns_ip: &str,
+    bootstrap_token: &str,
+    certificate: &str,
+    control_container_userdata: &str,
+    custom_user_data: &Option<CustomUserData>,
+) -> ProviderResult<String> {
+    let default_userdata = default_userdata(
+        endpoint,
+        cluster_dns_ip,
+        bootstrap_token,
+        certificate,
+        control_container_userdata,
+    );
+
+    let custom_userdata = if let Some(value) = custom_user_data {
+        value
+    } else {
+        return Ok(default_userdata);
+    };
+
+    match custom_userdata {
+        CustomUserData::Replace { encoded_userdata } => Ok(encoded_userdata.to_string()),
+        CustomUserData::Merge { encoded_userdata } => {
+            let merge_into = &mut decode_to_string(&default_userdata)?
+                .parse::<Value>()
+                .context(Resources::Clear, "Failed to parse TOML")?;
+            let merge_from = decode_to_string(encoded_userdata)?
+                .parse::<Value>()
+                .context(Resources::Clear, "Failed to parse TOML")?;
+            merge_values(&merge_from, merge_into)
+                .context(Resources::Clear, "Failed to merge TOML")?;
+            Ok(base64::encode(toml::to_string(merge_into).context(
+                Resources::Clear,
+                "Failed to serialize merged TOML",
+            )?))
+        }
+    }
+}
+
+fn default_userdata(
     endpoint: &str,
     cluster_dns_ip: &str,
     bootstrap_token: &str,
--- a/bottlerocket/agents/src/userdata.rs
+++ b/bottlerocket/agents/src/userdata.rs
@@ -1,0 +1,158 @@
+// Heavily borrowed from Bottlerocket's merge-toml crate.
+// See https://github.com/bottlerocket-os/bottlerocket/blob/v1.11.1/sources/api/storewolf/merge-toml/src/lib.rs
+
+use base64::decode;
+use resource_agent::provider::{IntoProviderError, ProviderError, ProviderResult, Resources};
+use std::str::from_utf8;
+use toml::{map::Entry, Value};
+
+/// This modifies the first given toml Value by inserting any values from the second Value.
+///
+/// This is done recursively.  Any time a scalar or array is seen, the left side is set to match
+/// the right side.  Any time a table is seen, we iterate through the keys of the tables; if the
+/// left side does not have the key from the right side, it's inserted, otherwise we recursively
+/// merge the values in each table for that key.
+///
+/// If at any point in the recursion the data types of the two values does not match, we error.
+pub fn merge_values<'a>(merge_from: &'a Value, merge_into: &'a mut Value) -> ProviderResult<()> {
+    // If the types of left and right don't match, we have inconsistent models, and shouldn't try
+    // to merge them.
+    if !merge_into.same_type(merge_from) {
+        IntoProviderError::context(
+            None,
+            Resources::Clear,
+            "Cannot merge mismatched data types in given TOML",
+        )?
+    }
+
+    match merge_from {
+        // If we see a scalar, we replace the left with the right.  We treat arrays like scalars so
+        // behavior is clear - no question about whether we're appending right onto left, etc.
+        Value::String(_)
+        | Value::Integer(_)
+        | Value::Float(_)
+        | Value::Boolean(_)
+        | Value::Datetime(_)
+        | Value::Array(_) => *merge_into = merge_from.clone(),
+
+        // If we see a table, we recursively merge each key.
+        Value::Table(from) => {
+            // We know the other side is a table because of the `ensure` above.
+            let to = merge_into.as_table_mut().ok_or_else(|| {
+                ProviderError::new_with_context(
+                    Resources::Clear,
+                    "Cannot merge mismatched data types in given TOML",
+                )
+            })?;
+            for (k_from, v_from) in from.iter() {
+                // Check if the left has the same key as the right.
+                match to.entry(k_from) {
+                    // If not, we can just insert the value.
+                    Entry::Vacant(e) => {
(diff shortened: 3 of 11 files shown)


bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -340,35 +344,52 @@ fn userdata(
...
-            r#"[settings.ecs]
-cluster = "{}""#,
-            cluster_name,
-        )),
-    })
+    let userdata = match custom_user_data {
+        Some(CustomUserData::Replace { encoded_userdata }) => encoded_userdata.to_string(),
+        _ => {
+            let temp_userdata = match cluster_type {
+                ClusterType::Eks => base64::encode(format!(
```
webern: Can you move these string formats out into two functions so that the branching logic here is easier to read?

```rust
fn default_eks_userdata(x, y, z) -> String { todo!() }
fn default_ecs_userdata(x) -> String { todo!() }
```

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -340,35 +344,52 @@ fn userdata(
...
-                .as_ref()
-                .context(memo, "Cluster DNS IP is required for eks clusters.")?,
-        )),
-        ClusterType::Ecs => base64::encode(format!(
-            r#"[settings.ecs]
-cluster = "{}""#,
-            cluster_name,
-        )),
-    })
+    let userdata = match custom_user_data {
```
webern: I would simplify this match case by early returning if there is no custom userdata. Something like 

```rust
let default_userdata = match cluster_type {
    ClusterType::Eks => default_eks_userdata(x, y, z),
    ClusterType::Ecs => default_ecs_userdata(x),
}

let custom_userdata = if let Some(value) = customer_user_data {
    value
} else {
    return Ok(default_userdata);
};

match custom_user_data {
    /* ... */
}
```

Note: in the case of `Replace` the `default_userdata` will have been constructed unnecessarily, but the easy-to-read program flow outweighs the construction of an unused string here.

bottlerocket/agents/src/bin/vsphere-vm-resource-agent/vsphere_vm_provider.rs
```diff
@@ -407,22 +411,39 @@ fn userdata(
...
-[settings.kubernetes]
-api-server = "https://{}:6443"
-cluster-dns-ip = "{}"
-bootstrap-token = "{}"
-cluster-certificate = "{}""#,
-        control_container_userdata, endpoint, cluster_dns_ip, bootstrap_token, certificate
-    ))
+    custom_user_data: &Option<CustomUserData>,
+) -> ProviderResult<String> {
+    let userdata = match custom_user_data {
```
webern: We could use the same program flow and refactor here that I described for EKS.

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// This modifies the first given toml Value by inserting any values from the second Value.
+///
+/// This is done recursively.  Any time a scalar or array is seen, the left side is set to match
+/// the right side.  Any time a table is seen, we iterate through the keys of the tables; if the
+/// left side does not have the key from the right side, it's inserted, otherwise we recursively
+/// merge the values in each table for that key.
+///
+/// If at any point in the recursion the data types of the two values does not match, we error.
+pub fn merge_values<'a>(merge_into: &'a mut Value, merge_from: &'a Value) -> ProviderResult<()> {
```
webern: I prefer left-to-right, `from, to` ordering (I always mess up when creating symlinks `ln -s to from`)

webern (commented):
I see you are pushing changes so I will save what I've commented on so far. Nice work.

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,106 @@
...
+        | Value::Integer(_)
+        | Value::Float(_)
+        | Value::Boolean(_)
+        | Value::Datetime(_)
+        | Value::Array(_) => *merge_into = merge_from.clone(),
+
+        // If we see a table, we recursively merge each key.
+        Value::Table(t2) => {
+            // We know the other side is a table because of the `ensure` above.
+            let t1 = merge_into.as_table_mut().unwrap();
```
webern: it's  safest to check again in case somebody deletes the code above. It's easiest to just say "the code must never panic" than to try to reason about `unwrap` statements. (then again, this isn't avionics, so :man_shrugging: )

```suggestion
            let t1 = merge_into.as_table_mut().ok_or_else(|| IntoProviderError::context(
            None,
            Resources::Clear,
            "Cannot merge mismatched data types in given TOML",
        ))?;
```

webern: Easier to read if we know which one is from/to. 

```suggestion
        Value::Table(from) => {
            // We know the other side is a table because of the `ensure` above.
            let to = merge_into.as_table_mut().unwrap();
```

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
+
+pub fn parse_toml(parse_string: String) -> ProviderResult<Value> {
+    IntoProviderError::context(
```
webern: Woah is this my fault? Should this be named IntoProviderResult instead? (don't change it in this PR, I am just surprised by this function body)

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,106 @@
...
+        parse_string.parse::<Value>(),
+        Resources::Clear,
+        "Failed to parse TOML",
+    )
+}
+
+#[cfg(test)]
+mod test {
```
webern, line 76: Thank you for writing a test. You could probably add a couple more to cover cases such as
- left has the key, right does not
- right has the key, left does not
- error cases

bottlerocket/testsys/src/run_aws_ecs.rs
```diff
@@ -151,6 +152,15 @@ pub(crate) struct RunAwsEcs {
...
+
+    /// The path to a TOML file containing custom userdata.
+    #[structopt(long, requires("custom_user_data_mode"))]
+    custom_user_data: Option<String>,
+
+    /// The way custom userdata should interact with the default userdata.
+    /// The possible values are `merge` and `replace`.
+    #[structopt(long, requires("custom_user_data"))]
+    custom_user_data_mode: Option<String>,
 }
```
webern: We shouldn't add two optional fields to a struct that must both be `Some` in order for the structure to be in a consistent state. Instead we should have

```rust
struct CustomUserDataConfig {
    data: String,
    mode: CustomUserDataMode,
}


```

Then in the main struct:

```
    /// The path to a TOML file containing custom userdata.
    #[structopt(long, requires("custom_user_data_mode"))]
    custom_user_data: Option<CustomUserData>,
```

I know @ecpullen doesn't like this "tagging"-patterned use of an enum, but in this case since it has to work with yaml it might be the only way to make it self-consistent and self-documenting that both fields are required.

bottlerocket/testsys/src/run_aws_ecs.rs
```diff
@@ -350,6 +360,27 @@ impl RunAwsEcs {
...
+                Ok(data) => match self.custom_user_data_mode.as_deref() {
+                    Some("merge") => Ok(Some(CustomUserData::Merge {
+                        encoded_userdata: data,
+                    })),
+                    Some("replace") => Ok(Some(CustomUserData::Replace {
+                        encoded_userdata: data,
+                    })),
+                    _ => Err(Error::InvalidArguments {
+                        why: "invalid custom userdata mode".to_string(),
+                    }),
```
webern: This is the error we should avoid. Especially since the "invalid custom userdata mode" might be `Option::None`.

If you add a "tagging" enum like I requested above, which I called `CustomerUserdataMode`, you could:

```rust
impl CustomUserData {
    fn new(x: CustomUserDataConfig) -> Self {
       match x.mode {
           /* ... */
       }
   }
}
```

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -171,6 +172,15 @@ pub(crate) struct RunAwsK8s {
...
     assume_role: Option<String>,
+
+    /// The path to a TOML file containing custom userdata.
+    #[structopt(long, requires("custom_user_data_mode"))]
+    custom_user_data: Option<String>,
+
+    /// The way custom userdata should interact with the default userdata.
+    /// The possible values are `merge` and `replace`.
+    #[structopt(long, requires("custom_user_data"))]
+    custom_user_data_mode: Option<String>,
```
webern: same thing, avoid designing structures that are not self-validating. i.e. struct.field_a should not rely on struct.field_b being in a compatible state.

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -369,6 +379,27 @@ impl RunAwsK8s {
         secrets: Option<BTreeMap<String, SecretName>>,
         cluster_resource_name: &str,
     ) -> Result<Resource> {
+        let user_data = if let Some(path) = &self.custom_user_data {
```
webern: repeated code. maybe my type suggestions above would eliminate it.

bottlerocket/testsys/src/run_vmware.rs
```diff
@@ -168,6 +169,15 @@ pub(crate) struct RunVmware {
...
     assume_role: Option<String>,
+
+    /// The path to a TOML file containing custom userdata.
+    #[structopt(long, requires("custom_user_data_mode"))]
+    custom_user_data: Option<String>,
+
+    /// The way custom userdata should interact with the default userdata.
+    /// The possible values are `merge` and `replace`.
+    #[structopt(long, requires("custom_user_data"))]
+    custom_user_data_mode: Option<String>,
```
webern: same comment about struct fields not relying on the state of other struct fields being consistent

bottlerocket/testsys/src/run_vmware.rs
```diff
@@ -317,6 +327,27 @@ impl RunVmware {
         secrets: Option<BTreeMap<String, SecretName>>,
         encoded_kubeconfig: &str,
     ) -> Result<Resource> {
+        let user_data = if let Some(path) = &self.custom_user_data {
```
webern: looks like repeated code,

mjsterckx:
^ Addressed some of webern's comments.

bottlerocket/testsys/src/error.rs
```diff
@@ -109,6 +109,9 @@ pub(crate) enum Error {
     #[snafu(display("Error patching {}: {}", what, source))]
     Patch { what: String, source: kube::Error },
 
+    #[snafu(display("Error parsing TOML file: {}", path))]
```
webern: Throughout our projects we format errors like this:
- When adding some input information we surround it with single quotes
- When including an underlying error we put it at the end after a colon, without single quotes.

So your error should be

```suggestion
    #[snafu(display("Error parsing TOML file '{}'", path))]
```
Ideally you would have an underlying error for context and it would look like this:

```
    #[snafu(display("Error parsing TOML file '{}': {}", path, source))]
```

webern (commented):
I see you are pushing changes so I will save what I've commented on so far. Nice work.

bottlerocket/testsys/src/run_aws_ecs.rs
mjsterckx: Doesn't the `requires` `structopt` ensure that either none of them are present, or both of them are?

That's how it's used in `cli` anyway.

mjsterckx: That's how it's used in `cli` anyway.

mjsterckx:
^ Added `source` field to `ParseToml` error type

bottlerocket/testsys/src/run_aws_ecs.rs
webern: > Doesn't the requires structopt 

It might, but that doesn't prevent you from writing gross code when you go to use the fields and have to handle the inconsistent-state cases anyway (because they are Options and this is Rust)...

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -335,35 +339,69 @@ fn userdata(
...
+    };
+
+    let custom_userdata = if let Some(value) = custom_user_data {
+        value
+    } else {
+        return Ok(default_userdata);
+    };
+
+    match custom_userdata {
+        CustomUserData::Replace { encoded_userdata } => Ok(encoded_userdata.to_string()),
```
ecpullen, line 360: Does the user data need to be decoded?

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -335,35 +339,69 @@ fn userdata(
...
+    endpoint: &Option<String>,
+    certificate: &Option<String>,
+    cluster_dns_ip: &Option<String>,
+    memo: &ProductionMemo,
+) -> Result<String, ProviderError> {
+    Ok(base64::encode(format!(
+        r#"[settings.updates]
 ignore-waves = true
-    
+
```
ecpullen, line 388: Is there a reason not to keep this newline?

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,106 @@
+use resource_agent::provider::{IntoProviderError, ProviderResult, Resources};
```
ecpullen: I would prefer to use the existing lib in Bottlerocket instead of duplicating all of the code.

bottlerocket/testsys/src/run_aws_ecs.rs
ecpullen: `testsys` is going away, so I ok with any solution (even leaving it out).

webern: Sorry it took me a minute to realize this was an Args struct. You can ignore this comment thread.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
mjsterckx: The newline is still there, I just removed a tab that must've been accidentally added at some point.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
mjsterckx: The TOML file is read as string and contained in `encoded_userdata` so I don't think so? But I suppose I should change the name of that variable.

ecpullen: The userdata should be base64 encoded.

mjsterckx:
^ Addressed several comments: switched `from`, `to` order in `merge_values`; replaced `unwrap` in `merge_values` with a second check to see if the types match; renamed `t2` and `t1` to `from` and `to` respectively; added code to the CLI to encode userdata as base64 and added a method to `userdata.rs` to decode back to a String.

bottlerocket/agents/src/userdata.rs
mjsterckx: `IntoProviderResult` does not exist (`IntoProviderError` does, but that's another matter). This function was removed, however, so I am marking this conversation as resolved.

mjsterckx:
^ Added a couple more unit tests (key present on left side but not on right; mismatched type error)

mjsterckx:
^ Small fixes related to encoding/decoding

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,158 @@
+// This is only used in the build script for now, but it's in a separate crate so we can have unit
+// tests.
```
etungsten:
```suggestion
// This is only used in the build script for now, but it's in a separate crate so we can have unit
// tests.
```
I don't think this comment applies here.

After looking through this, I think we should instead switch to using the `merge-toml` library from the bottlerocket repo directly https://github.com/bottlerocket-os/bottlerocket/blob/67e5abfb0e1575fb6bcb7a710cdb3d422ef12283/sources/api/storewolf/merge-toml/src/lib.rs like @ecpullen mentioned before. We can still use the custom error types and keep `decode_to_string` somewhere else/

etungsten (changes requested):

bottlerocket/agents/src/userdata.rs
mjsterckx: I can remove the comment, but copying the code instead of using the library is something that was discussed and @webern especially felt like we should not rely on a library from a different project that was never meant to be included in other projects.

webern: This is true. I am hesitant to create the first ever external dependency on library code that resides in Bottlerocket. I'd rather we take an action item to create a merge_toml crate external to both projects and put it on crates.io than to have dependencies going in both directions (Bottlerocket depends on testsys which depends on Bottlerocket).

etungsten: Can you add a comment to add some attribution for this code? Just to keep track that it's copied from bottlerocket. Adding the permalink and saying this is directly borrow from there is probably enough. https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/storewolf/merge-toml/src/lib.rs

See https://github.com/bottlerocket-os/bottlerocket/blob/f915de1a074d88a16348b444d31c49aaff51f8e9/sources/host-ctr/cmd/host-ctr/registry.go#L48-L52 for an example.

ecpullen: Can you also create an issue in both repos so we make sure this gets fixed.

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -171,6 +173,15 @@ pub(crate) struct RunAwsK8s {
...
     assume_role: Option<String>,
+
+    /// The path to a TOML file containing custom userdata.
+    #[structopt(long, requires("custom-user-data-mode"))]
+    custom_user_data: Option<String>,
+
+    /// The way custom userdata should interact with the default userdata.
+    /// The possible values are `merge` and `replace`.
+    #[structopt(long, requires("custom-user-data"))]
+    custom_user_data_mode: Option<String>,
```
ecpullen: Let's make this an enum.

bottlerocket/testsys/src/run_aws_k8s.rs
```diff
@@ -369,6 +380,28 @@ impl RunAwsK8s {
...
+                    }),
+                },
+                Err(err) => Err(Error::ParseToml {
+                    path: path.to_string(),
+                    source: err,
+                }),
+            }
+        } else {
+            Ok(None)
+        }?;
```
ecpullen:
```suggestion
let user_data = &self.custom_user_data
        .map(|path| read_to_string(path.clone()))
        .transpose()
        .context(error::ReadSnafu{})?
        .map(base64::encode);
        let user_data = match (self.custom_user_data_mode, user_data){
                  (Some(_),None)|(None,Some(_)) => return Err(error::Error::InvalidArguments { why: "Either both or neither of custom-user-data-mode and custom-user-data must be provided.".to__string() }),
                 (Some(UserdataType::Merge),Some(userdata)) => Some(CustomUserData::Merge{encoded_userdata: userdata}),
                 (Some(UserdataType::Replace),Some(userdata)) => Some(CustomUserData::Replace{encoded_userdata: userdata}),
                 (None,None) => None
        };
```

webern (approved):
Looks good when the erikson's and ethan's suggestions are implemented. Nice work.

mjsterckx:
^ Addressed ecpullen's suggestion and added attribution to indicate that `userdata.rs` is mostly copied from the Bottlerocket repo.

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,158 @@
+// Heavily borrowed from Bottlerocket's merge-toml crate.
+// See https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/storewolf/merge-toml/src/lib.rs
```
etungsten: quick nit: sorry, this should be a more permanent link that's tagged:
```suggestion
// See https://github.com/bottlerocket-os/bottlerocket/blob/v1.11.1/sources/api/storewolf/merge-toml/src/lib.rs
```

etungsten (approved):

etungsten (approved):

bottlerocket/agents/src/userdata.rs
```diff
@@ -0,0 +1,158 @@
...
+    Ok(())
+}
+
+pub fn decode_to_string(encoded_userdata: &String) -> ProviderResult<String> {
+    Ok(from_utf8(
+        &decode(encoded_userdata).context(Resources::Clear, "Failed to decode base64 TOML")?,
+    )
+    .context(Resources::Clear, "Failed to decode base64 TOML")?
+    .to_string())
+}
```
ecpullen, line 73: I'm not sure this belongs here, but I won't block on it.

ecpullen (approved):

bottlerocket/agents/src/userdata.rs
mjsterckx: For now, I think it's fine keeping it in a file called `userdata` since the function is only used in context of the userdata. Once the `merge-toml` code gets imported as library instead of being copied, this function can be moved to a `utils` location.

PR #684, opened 2022-12-05
TestManager, cli, Agent: Improvements to `status`

ecpullen:
**Issue number:**

Related to #679 
Closes #658 

**Description of changes:**

```
commit 75e173d2fa131244823cb337731cd2142a4f136d (HEAD -> testsys-status, origin/testsys-status)
Author: ecpullen <ecpullen@aol.com>
Date:   Mon Dec 5 18:41:17 2022 +0000

    TestManager: Don't truncate `Last Update`
    
    This sets the `MinWidth` for the `LAST UPDATE` column to 20 chars (the
    exact number required).

commit 318c88c5ccdb3430ead02e1295e087acff08b3e4
Author: ecpullen <ecpullen@aol.com>
Date:   Mon Dec 5 18:28:28 2022 +0000

    TestManager: Fix some columns out of order
    
    When `with_progress` and `with_time` were added, the order of crds was
    allowed to change. This caused the columns to be out of order. The crds
    are now resorted in each function.

commit cc26f55cb4a7a213054cf6f74776823648edb614
Author: ecpullen <ecpullen@aol.com>
Date:   Mon Dec 5 17:59:02 2022 +0000

    Controller, Agent: Report the last time of update
    
    Adds a new field to both `Test` and `Resource` status.
    All json patches now add `new_timestamp()` which adds the `SystemTime`
    to the `status/lastUpdate`.

commit ae036a83496dff45b02e97de419912a36775b2b5
Author: ecpullen <ecpullen@aol.com>
Date:   Fri Dec 2 18:26:44 2022 +0000

    TestManager: Additional column duplication
    
    This fixes a bug where a test with retries would cause additional column
    to not be aligned because the values were not duplicated for each result
    row.

commit d50efaac1d70f5802929553e2fdaee1b2a4225e7
Author: ecpullen <ecpullen@aol.com>
Date:   Fri Dec 2 18:07:39 2022 +0000

    TestManager: Add optional `STATUS` column
    
    A new optonal status column was added that displays the `currentStatus`
    of a resource or optionally the `other_info` of a test.
    `StatusProgress::Resources` will only show the status of resources while
    `StatusProgress::WithTests` will also include test status.
```

**Testing done:**

Tested with `cargo make test`

Initial status
```bash
$ cli status -pu
 NAME                                TYPE       STATE     PASSED   SKIPPED   FAILED  STATUS            LAST UPDATE           
 x86-64-aws-k8s-123                  Resource   unknown                                                 2022-12-05T18:47:44Z 
[...]

Issue #679, opened 2022-12-01
TestManager: Make `status` more useful


ecpullen:
The current `TestManager::status` provides an expandable table containing the crd names, state, and the number of passed and failed tests. This issue is to track potential enhancements and discuss other possible solutions.

- [x] #658 
- [x] #656 
- [x] Test updates while running
- [x] Show the status of the resource
- [x] Add more frequent updates for resource creation
  - [x] EKS resource provider
  - [x] EC2 resource provider
  - [x] vSphere k8s cluster resource provider
  - [x] vSphere vm resource provider
  - [x] ECS cluster provider
- [x] Regular status updates for tests
  - [x] sonobuoy
  - [x] workload

One potential improvement, is test results. While a test is running a user has no idea how many tests have passed, run, or still have to run. It would be nice if tests could send updates as a testing suite (`sonobuoy`) progresses.

Another place with room for improvement is resource creation. When a resource is running, the `status` reports running, and that's it. There is no insight to what the agent is doing. A field could be added to status that displays a brief explanation of what's happening in the agent. This would also provide insight to errors without having to read the agents logs.

If any of these proposals are worth looking into, I will create a separate issue for them.

ecpullen:
> One potential improvement, is test results. While a test is running a user has no idea how many tests have passed, run, or still have to run. It would be nice if tests could send updates as a testing suite (sonobuoy) progresses.


To solve this, we could have a `send_test_run_start` function  in the `Client` that would create a new item in the `TestResults` array. Then another function `send_updated_test_results` would replace the last item in the `TestResults` array. Whichever test result is the last in a running test would be the `running` test and results would be updated at whatever pace the agent sent them. With this proposal, the original flow of agents sending results at the end is unchanged, but it enables continuous status reporting for tests.

Another option is to replace `send_test_results` with the proposed `send_updated_test_result` and force all agents to follow that flow of information.

ecpullen:
> Another place with room for improvement is resource creation. When a resource is running, the status reports running, and that's it. There is no insight to what the agent is doing. A field could be added to status that displays a brief explanation of what's happening in the agent. This would also provide insight to errors without having to read the agents logs.

Most resource agents have a `current_status` field in their `ProductionMemo`. When creating the status table, the value for `current_status` could be added to a column. If a resource didn't have a `current_status` field, nothing would be printed.

webern:
[...]

PR #688, opened 2022-12-06
agents: Add more `current_status` updates for the `STATUS` column

ecpullen:
**Issue number:**

Related to #679 

**Description of changes:**

This is a follow up to #684 that adds more status reporting by the resource agents.

**Testing done:**

`make images`
`cargo make test`

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
+++ b/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
@@ -99,8 +99,22 @@ impl Create for Ec2Creator {
             .await
             .context(&memo, "Error storing uuid in info client")?;
 
+        info!("Getting AWS secret");
+        memo.current_status = "Getting AWS secret".to_string();
+        client
+            .send_info(memo.clone())
+            .await
+            .context(Resources::Clear, "Error sending cluster creation message")?;
+
         memo.aws_secret_name = spec.secrets.get(AWS_CREDENTIALS_SECRET_NAME).cloned();
         memo.assume_role = spec.configuration.assume_role.clone();
+
+        info!("Creating AWS config");
+        memo.current_status = "Creating AWS config".to_string();
+        client
+            .send_info(memo.clone())
+            .await
+            .context(Resources::Clear, "Error sending cluster creation message")?;
 
         let shared_config = aws_config(
             &spec.secrets.get(AWS_CREDENTIALS_SECRET_NAME),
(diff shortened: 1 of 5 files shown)


bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -113,14 +125,22 @@ impl Create for Ec2Creator {
...
 
-        let run_instance_result = run_instances(&spec, &ec2_client, instance_uuid, &memo).await?;
+        memo.current_status = "Launching Instance(s)".to_string();
+        client
+            .send_info(memo.clone())
+            .await
+            .context(Resources::Clear, "Error sending cluster creation message")?;
+
+        let run_instance_result =
+            run_instances(&spec, &ec2_client, instance_uuid, &mut memo, client).await?;
```
ecpullen, line 138: Not sure why this got formatted now and not before?

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -99,9 +99,21 @@ impl Create for Ec2Creator {
             .await
             .context(&memo, "Error storing uuid in info client")?;
 
+        memo.current_status = "Getting AWS Secret".to_string();
```
mjsterckx: Kind of a follow-up to my comment in https://github.com/bottlerocket-os/bottlerocket-test-system/pull/684: since the status there is reported with only a capital letter for the first word (e.g. `No test results`), could we do the same for these custom Strings to keep it consistent (obviously AWS stays capitalized but this would become `Getting AWS secret`)?

ecpullen: I prefer Caps for readability, but I am happy to change it to whatever we decide. I do agree that we should be consistent, though.

webern (approved):
Could we also log these as `info!` for troubleshooting (since the CRD will only have the last one that made it to the k8s api)?

ecpullen:
^ Changed status format, add `info!()` wherever the status was reported.

stmcginnis (approved):
Looks good to me.

Only thing that came to mind - wondering if we should just add a method to `memo` to handle some of this. We have this pattern repeated over and over:

```rust
info!("Launching worker nodes");
memo.current_status = "Launching worker nodes".to_string();
client.send_info(memo.clone())
```

It might be nice to streamline some of that and get rid of duplication by having something like:

```rust
fn update_status(&self, status: &str, client: &T) {
    info!(status);
    self.current_status = status.to_string();
    client.send_info(self.clone())
}
```

Just something to think about. I think how it is now is good.

etungsten (approved):

ecpullen:
I was thinking that too. We could in the future add a send_status function to the client that would put the message in the `current_status` field of the crd (even if the Memo didn't have that field).

PR #690, opened 2022-12-07
Agents: Allow continuous updates by test agents

ecpullen:
**Issue number:**

Related #679 

**Description of changes:**

Adds a new client, `InfoClient` that allows a test to update `status.current_test`, a new field in the test's status. `current_test` represents the results for the current test that is being run and is given its own row in status reporting.

**Testing done:**

Tested with the `example-resource-agent` and saw the status change for each iteration of the loop and `LAST UPDATE` is updated every 3 seconds (the `hello_duration` in my yaml file).

**Terms of contribution:**

### DIFF
--- a/model/src/test.rs
+++ b/model/src/test.rs
@@ -5,6 +5,7 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use kube::CustomResource;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use serde_plain::derive_display_from_serialize;
 use std::borrow::Cow;
 
 /// A TestSys Test. The `CustomResource` derive also produces a struct named `Test` which represents
@@ -59,7 +60,10 @@ pub enum Outcome {
     Fail,
     Timeout,
     Unknown,
+    InProgress,
 }
+
+derive_display_from_serialize!(Outcome);
 
 impl Default for Outcome {
     fn default() -> Self {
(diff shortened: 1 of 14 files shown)


ecpullen:
^ Add workload test, fix commit order.

model/src/test_manager/status.rs
```diff
@@ -310,7 +323,7 @@ impl From<&Crd> for Vec<ResultRow> {
                         results.push(ResultRow {
                             name: retry_name,
                             object_type: "Test".to_string(),
-                            state: state.clone(),
+                            state: result.outcome.to_string(),
```
ecpullen, line 326: The test outcome is more useful than the test state.

model/src/test.rs
```diff
@@ -182,7 +187,7 @@ impl Test {
                         Outcome::Pass => TestUserState::Passed,
                         Outcome::Fail => TestUserState::Failed,
                         Outcome::Timeout => TestUserState::Failed,
-                        Outcome::Unknown => {
+                        Outcome::Unknown | Outcome::InProgress => {
```
webern, line 190: I don't remember why we were returning "Passed" for a test with Outcome::Unknown, do you? Seems a little weird to say an InProgress test has "Passed", but I don't know the consequences or how that might be encountered.

webern (approved):
You split this up such that a follow up PR will be the one that uses the new functionality, right? Looks good to me. Seems like something we had considered a while back.

model/src/test.rs
ecpullen: I think the original logic was that we check to make sure there are no failed tests.

stmcginnis (approved):
Looks good to me.

etungsten (approved):

Issue #656, opened 2022-11-11
TestManager: Better status filtering


ecpullen:
If many tests are being run, there will be many columns in testsys status and not all are visible when using cargo make watch-test. A set of filtering methods should be implemented for the status command to limit status results.

Ideas:
* Crd type (Test, Resource)
* Build
* Running tests
* Finished tests
* Failed tests
* Unfinished tests

These new filters should probably be incorporated in the form of `SelectionParams`

webern:
Maybe... we should think about whether the same could be done with `grep` though. I know this was a request from @bcressey but sometimes the best solution is a simple one, so I want to make sure we don't build grep inside of testsys.

ecpullen:
This is a bigger picture issue than just status filtering. By expanding the `SelectionParams`, we can offer different delete methods as well i.e delete all passed tests.

PR #695, opened 2022-12-09
cli: Improve status filtering

ecpullen:
**Issue number:**

#656 

**Description of changes:**

Even though most of this change could be accomplished with `grep`, the improvement to `SelectionParams` will give users a much better control of deletion (in future pr).

    cli: Improve status filtering
    
    Convert `SelectionParams` from an enum to a struct.
    Add filtering based on the state of the objects.
    - Running
    - Completed
    - Passed
    - Failed
    - NotFinished
    
    Add filtering based on the crd type.
    - Tests
    - Resource
    
    Add new filtering to `cli status`
    - '-t' Only show tests
    - '-r' Only show resources
    - '--labels foo=bar' Filter based on labels in the CRD
    - '--name' Only show tests/resources with the specified name
    - '--state' Filter tests/resources by the given state

```bash
$ cli status -h
Get the status of testsys objects

Usage: cli status [OPTIONS]

Options:
      --json             Output the results in JSON format
  -c, --controller       Check the status of the testsys controller
  -p, --progress         Include the status of resources when reporting status
      --with-test        Include the `Test` status, too. Requires `--progress`
  -u, --with-time        Include the time the CRD was last updated
  -t, --tests            Include `Test`s (if passed with `--resources`, `Test`s and `Resource`s will be shown)
  -r, --resources        Include `Resource`s (if passed with `--tests`, `Test`s and `Resource`s will be shown)
      --labels <LABELS>  Only include objects with the specified labels ("foo=bar,biz=baz")
      --state <STATE>    Only include objects with the specified state ("completed", "running", "not-finished", "passed", "failed")
      --name <NAME>      Only include objects with the specified name
  -h, --help             Print help information
```

**Testing done:**

Tested various combinations of cli invocations and ensured that the status reported represented the cli query.

**Terms of contribution:**

### DIFF
--- a/cli/src/status.rs
+++ b/cli/src/status.rs
@@ -1,6 +1,6 @@
 use anyhow::{Context, Result};
 use clap::Parser;
-use model::test_manager::{StatusProgress, TestManager};
+use model::test_manager::{CrdState, CrdType, SelectionParams, StatusProgress, TestManager};
 use terminal_size::{Height, Width};
 
 /// Check the status of a TestSys object.
@@ -19,18 +19,50 @@ pub(crate) struct Status {
     progress: bool,
 
     /// Include the `Test` status, too. Requires `--progress`
-    #[clap(long, short = 't', requires("progress"))]
+    #[clap(long, requires("progress"))]
     with_test: bool,
 
     /// Include the time the CRD was last updated
     #[clap(long, short = 'u')]
     with_time: bool,
+
+    /// Include `Test`s (if passed with `--resources`, `Test`s and `Resource`s will be shown)
+    #[clap(long, short = 't')]
+    tests: bool,
+
+    /// Include `Resource`s (if passed with `--tests`, `Test`s and `Resource`s will be shown)
+    #[clap(long, short = 'r')]
+    resources: bool,
+
+    /// Only include objects with the specified labels ("foo=bar,biz=baz")
+    #[clap(long)]
+    labels: Option<String>,
+
+    /// Only include objects with the specified state ("completed", "running", "not-finished",
+    /// "passed", "failed")
+    #[clap(long)]
+    state: Option<CrdState>,
+
+    /// Only include objects with the specified name
+    #[clap(long)]
+    name: Option<String>,
 }
 
 impl Status {
     pub(crate) async fn run(self, client: TestManager) -> Result<()> {
+        let crd_type = match (self.tests, self.resources) {
+            (true, false) => Some(CrdType::Test),
+            (false, true) => Some(CrdType::Resource),
+            _ => None,
+        };
+        let selection_params = SelectionParams {
+            crd_type,
+            labels: self.labels,
+            name: self.name,
+            state: self.state,
+        };
         let mut status = client
-            .status(&Default::default(), self.controller)
+            .status(&selection_params, self.controller)
             .await
             .context("Unable to get status")?;
 
--- a/model/src/test_manager/manager.rs
+++ b/model/src/test_manager/manager.rs
@@ -1,11 +1,11 @@
 use super::{
-    error, DeleteEvent, DockerConfigJson, ImageConfig, ResourceState, Result, SelectionParams,
-    StatusSnapshot,
+    error, CrdState, CrdType, DeleteEvent, DockerConfigJson, ImageConfig, ResourceState, Result,
+    SelectionParams, StatusSnapshot,
 };
 use crate::clients::{CrdClient, ResourceClient, TestClient};
 use crate::constants::{LABEL_COMPONENT, TESTSYS_RESULTS_FILE};
 use crate::system::AgentType;
-use crate::{Crd, CrdName, SecretName};
+use crate::{Crd, CrdName, Resource, SecretName, TaskState, Test, TestUserState};
 use bytes::Bytes;
 use futures::{Stream, StreamExt};
 use k8s_openapi::api::core::v1::{Pod, Secret};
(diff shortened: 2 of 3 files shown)


mjsterckx:
Does the `--name` option support partial matching? For example, if I wanted to see all k8s tests while there are several ecs and other tests in the cluster?

ecpullen:
> Does the `--name` option support partial matching? For example, if I wanted to see all k8s tests while there are several ecs and other tests in the cluster?

It does not. We could look into adding that in the future.

mjsterckx (approved):
I would like partial matching on the `name` field, but that is not a blocker. Everything else looks great.

cli/src/status.rs
```diff
@@ -19,18 +19,49 @@ pub(crate) struct Status {
...
     #[clap(long, short = 'u')]
     with_time: bool,
+
+    /// Only include `Test`s (if passed with `--resources`, `Test`s and `Resource`s will be shown)
+    #[clap(long, short = 't')]
+    tests: bool,
+
+    /// Only include `Resource`s (if passed with `--tests`, `Test`s and `Resource`s will be shown)
+    #[clap(long, short = 'r')]
+    resources: bool,
```
webern: What happens when I pass `--tests --resources`, etc? Use of the word `only` implies that this would return nothing.

cli/src/status.rs
```diff
@@ -19,18 +19,49 @@ pub(crate) struct Status {
...
+    /// Only include objects with the specified name
+    #[clap(long)]
+    name: Option<String>,
 }
 
 impl Status {
     pub(crate) async fn run(self, client: TestManager) -> Result<()> {
+        let crd_type = match (self.tests, self.resources) {
+            (true, false) => Some(CrdType::Test),
+            (false, true) => Some(CrdType::Resource),
```
webern, line 55: ~~Maybe `--tests` and `--resources` should be exclusive in `clap` to be more clear to the user. This is kind of surprising.~~

Not surprising, I misread it.

cli/src/status.rs
```diff
@@ -19,18 +19,49 @@ pub(crate) struct Status {
...
+
+    /// Only include `Resource`s (if passed with `--tests`, `Test`s and `Resource`s will be shown)
+    #[clap(long, short = 'r')]
+    resources: bool,
+
+    /// Only include objects with the specified labels ("foo=bar,biz=baz")
+    #[clap(long)]
+    labels: Option<String>,
+
+    /// Only include objects with the specified state
```
webern: List the allowable strings here. I can't tell you how frustrating it is to get a sentence like this in a CLI and have to go searching the internet to find out what values I can pass.

model/src/test_manager/manager.rs
```diff
@@ -186,68 +186,112 @@ impl TestManager {
...
+                                            | TestUserState::ResourceError
+                                    )
+                                }
+                                CrdState::NotFinished => matches!(
+                                    test.test_user_state(),
+                                    TestUserState::Running
+                                        | TestUserState::Waiting
+                                        | TestUserState::Unknown
+                                ),
+                            }
```
webern: Can you create a named `-> bool` function for this match statement?

model/src/test_manager/manager.rs
```diff
@@ -186,68 +186,112 @@ impl TestManager {
...
+                                CrdState::NotFinished => {
+                                    matches!(
+                                        test.creation_task_state(),
+                                        TaskState::Running | TaskState::Unknown
+                                    ) || matches!(
+                                        test.creation_task_state(),
+                                        TaskState::Running | TaskState::Unknown
+                                    )
+                                }
+                                _ => false,
```
webern: Same here, busting out a named `-> bool` function would help readability.

model/src/test_manager/mod.rs
```diff
@@ -14,20 +13,45 @@ mod manager;
...
+    /// All `Test`s and `Resource`s that are not finished.
+    NotFinished,
+    /// All `Test`s and `Resource`s that are currently running.
+    Running,
+    /// All `Test`s and `Resource`s that are finished.
+    Completed,
+    /// All `Test`s that passed.
+    Passed,
+    /// All `Test`s that failed.
+    Failed,
```
webern, line 50: Hmmm, what does it mean if I say `--resources --state passed`?

webern (commented):
I realize this is the "deprecated" CLI, however it feels like a proving ground for what we will do in the Bottlerocket repo, hence my comments.

cli/src/status.rs
ecpullen: (if passed with `--tests`, `Test`s and `Resource`s will be shown

cli/src/status.rs
ecpullen: To me it makes a sense that if a user requests `--test` results then tests are shown, `--resource` => resources are shown then `--test` and `--resource` => tests and resources would be shown. While specifying nothing is equivalent to all.

model/src/test_manager/mod.rs
ecpullen: It will filter out non-resources and will show all of the tests that passed. (Nothing will be shown).

webern: > It will filter out non-resources and will show all of the tests that passed. (Nothing will be shown).

Is that the best design? What if we made it more generally applicable to Resources and Tests by eliminating the ability to list both passed and failed at the same time...

The issue I'm trying to solve for is a CLI that can take nonsensical input.

```rust
pub enum CrdState {
    /// All `Test`s and `Resource`s that are not finished.
    NotFinished,
    /// All `Test`s and `Resource`s that are currently running.
    Running,
    /// All `Test`s  that passed and `Resource`s that have been successfully created.
    Succeeded,
    /// All `Test`s that failed and `Resource`s that have failed to be created.
    Failed,
}
```

ecpullen: I do think that is the best design for the user.
I want a user to be able to call `cli status --state passed` and only see the passed tests. The user shouldn't care about the state of the resources. Same with `cli status --state failed`.We shouldn't require `cli status --test --state passed` for the main use case. IMO

webern (changes requested):
Blocking for now because I don't think it's good user interface design to accept user input that is effectively bad and return no results. Will come back and revisit this after discussion, etc.

`--resources --state passed` will do nothing... this is because the `--state` argument accepts two values that only make sense with `--tests` is present.

model/src/test_manager/mod.rs
ecpullen: In case there is confusion, `--state passed` does not require that `--tests` be present. `cli status` will print out all crds. By adding the `--test`/`--resource` flag, only those crds will be shown. By specifying a `state`, The output from `cli status` with the `--test` or `--resource` will get filtered to those matching the state specified. We can error out if `--state passed --resources` is used, but that seems unnecessary.

webern (approved):
> In case there is confusion, --state passed does not require that --tests be present. cli status will print out all crds. By adding the --test/--resource flag, only those crds will be shown. By specifying a state, The output from cli status with the --test or --resource will get filtered to those matching the state specified. We can error out if --state passed --resources is used, but that seems unnecessary.

Thank you for clarifying. In thinking about this further the only alternative to the shared enum that I can see would be `--test-state passed` `--resource-state completed`. It's an idea, but not good enough to block on.

The only blocking edit for me is to make sure the CLI `--help` prints the allowable strings for arguments that take an enumeration. With that change LGTM.

Thank you!

cli/src/status.rs
etungsten: Can we get rid of the "Only" in the comments for both `--test` and `--resource`? I share @webern's sentiment that this is a bit confusing.

etungsten (commented):
Other than the outstanding comments, LGTM.

ecpullen:
^ Moves matches in list to separate function
Remove `only` from test and resource help
Add acceptable `--state` values

stmcginnis (approved):

etungsten (approved):

Issue #419, opened 2022-05-20
workload test agent


webern:
Create a test agent for running container workloads on Bottlerocket. This test agent should take a generic configuration that is appropriate for both ECS and K8S so that we can run the same workload on `aws-ecs-*` and `*-k8s-*` variants of Bottlerocket.

When targeting a k8s variant, the agent will:
- create a sonobuoy plugin configuration that will run the workload
- within the workload, #418 will write test result information in a sonobuoy appropriate way
- the test agent will collect and report the results using the sonobuoy binary

When targeting an ecs variant, the agent will:
- create and run the workload as an ECS task
- within the workload, #418 will write test result information in a way that we want
- the test agent will collect and report the results in a bespoke manner

- [x] #616 
- [x] #617

depends on:
- #418

Here is some writing I did elsewhere to communicate the idea of a "workload test":

---

### Workload Tests

There is another class of test which we will call *Workload Tests*. Workload tests run a pod or task on Bottlerocket by way of the container orchestrator. Currently, we test Bottlerocket with Sonobuoy’s e2e conformance test suite. This is a workload test. 

For a workload test, the TestSys Test Agent launches a pod or task into the target cluster (the cluster where Bottlerocket nodes to be tested are running). The test code will run on a Bottlerocket node in the target cluster. For example, the NVIDIA smoke test will be launched into a K8s or ECS cluster where it will check that Tensorflow is working with GPU support. Essentially the same test will be valid for aws-ecs-1-nvidia as aws-k8s-1.21-nvidia. It is only the language with which the pod/task is specified that differs.

# Workload Tests for Bottlerocket

Workload tests run K8s pods or ECS task on Bottlerocket instances. They help ensure that the container orchestrator (kubelet or ecs-agent) is able to do its thing properly on Bottlerocket. Currently, we run Sonobuoy's "e2e" "plugin" to ensure k8s variance are compliant. We run an test suite to qualify the ECS variant, this is also a workload test.

We run sonobuoy "e2e" in "quick" mode and we run a simple ECS task for upgrade downgrade testing. These are *workload* tests. (Interestingly the actual process of updating the instance is modeled currently as an OS test.) We propose modeling Bottlerocket workload tests as custom sonobuoy plugins for K8s, and making a similar mechanism for ECS.

[...]

PR #696, opened 2022-12-09
Add NVIDIA workload test definition

stmcginnis:
**Issue number:**

Related: #419

**Description of changes:**

This adds a new `tests` directory tree to contain our workload tests and any other type of add-on test type we decide to add in the future. It adds our first workload test for running NVIDIA smoketests. Basic instructions are included to help guide future test authors of the requirements for tests and where they should be placed.

Some of this content will need to be updated once we fully finish publishing the container images and flesh out things like ECS testing and how the full execution of workload tests will be performed. This gets the majority of the pieces in place so those updates will be smaller and it will be easier to complete that work using published content.

**Testing done:**

Most testing was done when the initial `k8s-workload-agent` was added. The `nvidia-smoke` container creation has been tested by building the multiarch image and running it manually on both amd64 and arm64 ECS Bottlerocket nodes and viewing the output results.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/tests/workload/README.md
+++ b/bottlerocket/tests/workload/README.md
@@ -1,0 +1,78 @@
+# TestSys Workload Tests
+
+This directory contains the source files and artifacts for running TestSys workloads.
+
+## Workload Tests
+
+New workload tests should be added under this subdirectory.
+The current list of tests are detailed below.
+
+**NVIDIA smoketests**
+
+This workload tests executes various CUDA samples to verify GPU functionality for Bottlerocket hosts running on NVIDIA instances.
+See [nvidia-smoke.yaml](nvidia-smoke.yaml) for an example configuration.
+
+---
+
+## Workload Test Requirements
+
+Workload tests are container images that execute one or more application workloads.
+
+The expectation of test authors is to provide a container image with the following:
+
+* Provide a script or binary that will run in a container as part of a Kubernetes pod or ECS task
+* Provide a `./run.sh` entry point for the test container, optional arguments may be defined through environment variables
+* The shell script may be used directly to perform the test, or be a wrapper to allow executing other binaries or scripts
+* Use the `$RESULTS_DIR` environment variable to determine where to place output
+* At end of execution, collect test output into a results file `results.tar.gz` placed in `$RESULTS_DIR` with all relevant output
+* Write the `${RESULTS_DIR}/results.tar.gz` full path name to a `${RESULTS_DIR}/done` file
+* Container exit code determines pass/fail - 0 is considered successful completion
+
+These requirements will make the container compliant as a Sonobuoy plugin when run on Kubernetes.
+
+### Examples
+
+To help illustrate, the following could be used as the `run.sh` entry point script:
+
+```bash
+#!/bin/env bash
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+mkdir -p "${results_dir}"
+
+# Example of optional argument passed as environment variable to the container
+name="${NAME:-spam}"
+
+saveResults() {
+     cd "${results_dir}"
+     tar czf results.tar.gz *
+     echo "${results_dir}/results.tar.gz" > "${results_dir}/done"
+}
+
+# Make sure to always capture results in expected place and format
+trap saveResults EXIT
+
+# Run whatever you intest to test using this container
+/bin/tests/benchmark --name "${name}" >> "${results_dir}/${test_name}-results" 2>&1
+```
+
+The container image would then be defined with this script as its `ENTRYPOINT` and could be run by TestSys using something like:
+
+```yaml
+apiVersion: testsys.system/v1
+kind: Test
+metadata:
+  name: spam-workload
+  namespace: testsys
+spec:
+  agent:
+    configuration:
+      kubeconfigBase64: <Base64 encoded kubeconfig for the test cluster workload runs the tests in>
+      plugins:
+      - name: spam-workload
+        image: example/benchmark:v0.0.3
+    image: <your k8s-workload-agent image URI>
+    name: workload-test-agent
+    keepRunning: true
+  resources: {}
+```
--- a/bottlerocket/tests/workload/nvidia-smoke.yaml
+++ b/bottlerocket/tests/workload/nvidia-smoke.yaml
@@ -1,0 +1,16 @@
+kind: Test
+metadata:
+  name: workload-agent
+  namespace: testsys
+spec:
+  agent:
+    name: workload-agent
+    image: <WORKLOAD-AGENT-IMAGE>
+    keepRunning: false
+    configuration:
+      kubeconfigBase64: <BASE64-KUBECONFIG>
+      plugins:
+        - name: nvidia-workload
+          image: <NVIDIA-WORKLOAD-IMAGE>
+  resources: []
+  dependsOn: []
--- a/bottlerocket/tests/workload/nvidia-smoke/Makefile
+++ b/bottlerocket/tests/workload/nvidia-smoke/Makefile
@@ -1,0 +1,28 @@
+.DEFAULT_GOAL:=publish
+
+#### Variables
+PUBLISH_IMAGES_REGISTRY ?=
+TAG ?= latest
+REPOSITORY ?= gpu-tests
+
+#### Format the registry portion of the image tag, if needed
+REGISTRY ?=
+ifdef PUBLISH_IMAGES_REGISTRY
+REGISTRY = "$(PUBLISH_IMAGES_REGISTRY)/"
+endif
+
+#### Actual build targets
+build:
+ifndef REGISTRY
+$(error "The PUBLISH_IMAGES_REGISTRY value must be provided (e.g. PUBLISH_IMAGES_REGISTRY=861807767978.dkr.ecr.us-east-2.amazonaws.com make)")
+endif
+	docker buildx build --platform linux/arm64,linux/amd64 --tag $(REGISTRY)$(REPOSITORY):$(TAG) .
+
+publish:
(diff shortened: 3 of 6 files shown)


tests/workload/nvidia-smoke/Makefile
```diff
@@ -0,0 +1,25 @@
+.DEFAULT_GOAL:=publish
+
+#### Variables
+REGISTRY ?=
```
ecpullen: Should we use the same name as the root makefile `PUBLISH_IMAGES_REGISTRY`?

tests/workload/nvidia-smoke/Makefile
```diff
@@ -0,0 +1,25 @@
...
+TAG ?= latest
+REPOSITORY ?= gpu-tests
+
+#### Sanity check we have everything
+ifndef REGISTRY
+$(error "The REGISTRY value must be provided (e.g. REGISTRY=861807767978.dkr.ecr.us-east-2.amazonaws.com make)")
+endif
```
ecpullen: Is there a reason a user may not need to specify the `REGISTRY` i.e `kind load`

tests/workload/nvidia-smoke/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+```bash
+docker buildx inspect | grep Platforms
+```
+
+The output from this command should show both `linux/amd64` and `linux/arm64` platforms.
+Compiling the `amd64` binaries on an `arm64` host using emulation will be significantly slower.
+See below for instructions on how to set up a remote build host to offload this activity.
+
+This guide assumes you use ECR as your container's registry, and that you already have the ECR credentials plugin configured in your host.
```
ecpullen: Is it possible to write this guide without assuming `ECR`

ecpullen (commented):
Awesome!

stmcginnis:
I am now seeing the same failures on arm64 as I was seeing when trying to build on amd64:

```txt
 > [linux/amd64 builder  5/15] RUN cd /cuda-samples/Samples/0_Introduction/vectorAdd/ && make && cp $(basename $(pwd)) /samples/:
#0 0.200 Makefile:86: *** ERROR - unsupported value Error while loading /usr/local/nvidia/bin/uname: No such file or directory for TARGET_ARCH!.  Stop.
------
Dockerfile:21
--------------------
  19 |     # it filters out what samples won't be build.
  20 |     # This only includes the samples that should run on both amd64 and arm64.
  21 | >>> RUN cd /cuda-samples/Samples/0_Introduction/vectorAdd/ && make && cp $(basename $(pwd)) /samples/
  22 |     RUN cd /cuda-samples/Samples/0_Introduction/simpleVoteIntrinsics/ && make && cp $(basename $(pwd)) /samples/
  23 |     RUN cd /cuda-samples/Samples/0_Introduction/simpleAtomicIntrinsics/ && make && cp $(basename $(pwd)) /samples/
--------------------
ERROR: failed to solve: process "/dev/.buildkit_qemu_emulator /bin/sh -c cd /cuda-samples/Samples/0_Introduction/vectorAdd/ && make && cp $(basename $(pwd)) /samples/" did not complete successfully: exit code: 2
make: *** [Makefile:17: build] Error 1
```

Going to mark as draft until I understand the root cause and what has changed from the 50+ times I had built this successfully before submitting this PR.

tests/workload/nvidia-smoke/Makefile
stmcginnis: Yeah, that makes sense. I don't really like the length of that variable name, but it's better to have consistency than less typing.

tests/workload/nvidia-smoke/Makefile
stmcginnis: Probably not in the case of this particular workload test, but I see what you're saying. Since this is setting the pattern for future workload tests, I'll update this to check if the variable is set. If so, then it can format the `$REGISTRY/` prefix, and if not it can be empty.

tests/workload/nvidia-smoke/README.md
stmcginnis: Reworded this note since there really isn't anything that assumes or expects ECR vs any other image registry.

tests/workload/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+To help illustrate, the following could be used as the `run.sh` entry point script:
+
+```bash
+#!/bin/env bash
+
+results_dir="${RESULTS_DIR:-/tmp/results}"
+mkdir -p "${results_dir}"
+
+# Example of option argument passed as environment variable to the container
```
webern:
```suggestion
# Example of optional argument passed as environment variable to the container
```

stmcginnis:
Had to take a slightly different approach to be able to get a consistent good build. It has the added benefit that it is much, much faster due to not needing to compile under emulation, so the build completes in a fraction of the time from the first approach.

Addressed comments and moved the test tree under `bottlerocket/` since these are all Bottlerocket specific things.

bottlerocket/tests/workload/README.md
```diff
@@ -0,0 +1,77 @@
...
+saveResults() {
+     cd "${results_dir}"
+     tar czf results.tar.gz *
+     echo "${results_dir}/results.tar.gz" > "${results_dir}/done"
+}
+
+# Make sure to always capture results in expected place and format
+trap saveResults EXIT
+
+/bin/tests/benchmark --name "${name}" >> "${results_dir}/${test_name}-results" 2>&1
```
webern:
```suggestion
# Run whatever you intend to test using this container.
/bin/tests/benchmark --name "${name}" >> "${results_dir}/${test_name}-results" 2>&1
```

bottlerocket/tests/workload/nvidia-smoke.yaml
```diff
@@ -0,0 +1,16 @@
+kind: Test
+metadata:
+  name: workload-agent
+  namespace: testsys
+spec:
+  agent:
+    name: workload-agent
+    image: __WORKLOAD_AGENT_IMAGE__
```
webern: Is this a templating mechanism?

bottlerocket/tests/workload/nvidia-smoke/Makefile
```diff
@@ -0,0 +1,28 @@
...
+	docker buildx build --platform linux/arm64,linux/amd64 --tag $(REGISTRY)$(REPOSITORY):$(TAG) .
+
+publish:
+	docker buildx build --push --platform linux/arm64,linux/amd64 --tag $(REGISTRY)$(REPOSITORY):$(TAG) .
+
+x86_64:
+	docker buildx build --platform linux/amd64 --tag $(REGISTRY)$(REPOSITORY):$(TAG)-x86_64 .
+
+aarch64:
+	docker buildx build --platform linux/arm64 --tag $(REGISTRY)$(REPOSITORY):$(TAG)-aarch64 .
```
webern, line 28: Should we have a multiarch manifest instead?

bottlerocket/tests/workload/nvidia-smoke/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+In order to build the image for both architectures, you need to set up the `buildx` docker plugin.
+[This guide describes](https://github.com/docker/buildx#installing) the installation process.
+
+By default, the install of `docker buildx` uses the `docker` driver type.
+This driver is not able to build for all platform targets we need.
+You will need to run the following to create a new builder using the `docker-container` driver:
+
+```bash
+docker buildx create --use --bootstrap
```
webern, line 18: This environment set up and subsequent 15 steps make me sad. :cry: 


Do I understand correctly that you are running another machine for the target architecture? Is it common to do with qemu?
Not your fault though, cross-arch building of containers is what it is...

Thank you for documenting this well.

stmcginnis: Yeah, that's how I originally had it - with one host. It made cross-compiling incredibly slow.

And it was working for me running from an arm64 host fine, but when I tried running from an amd64 host I would get weird compile/linking errors with the NVIDIA sample binaries. At first I chalked this up to something odd with building for ARM on AMD, but then after cleaning up and resetting things I started to get the same errors on my arm64 host that had been previously working.

So I think in general it is fine to do both from one host, but in this particular case there is something about how the Makefiles do various checks and set variables that control compilation that somehow gets messed up when using the normal QEMU based route.

Unfortunately the only way I could find to reliably get things to build right was the currently documented approach of using two different machines. I'm not happy about it, but it's the best I could come up with.

It does have the nice benefit that compilation is much faster than the previous emulated approach - by probably 5x. So there is some setup pain to do it this way, but it kind of pays off in other ways.

bottlerocket/tests/workload/nvidia-smoke/Makefile
stmcginnis: These last two are really just convenience targets and could really be removed. They were useful when testing things out on different hosts. In practice, the publish target is really all that is needed and it does create a multiarch image.

bottlerocket/tests/workload/nvidia-smoke.yaml
stmcginnis: I was trying to strike a balance between showing an example that someone could read in the docs and understanding what they would need to put in a file, and having something that we could potentially do some sort of sed replacement to use it as a template and pull it into some of our testing workflow.

This could be changed to have some dummy value if you think this is too odd or confusing. We could also wait until the official images are published and update these with those values once they are available.

ecpullen: In most of the documentation I have written I have used `<SCREAMING-KEBAB-CASE>` to identify a value that should be replaced.

stmcginnis: I've changed this to match that pattern.

webern (approved):

ecpullen (approved):

PR #697, opened 2022-12-10
actions-workflow: use 16-core GitHub hosted runners

etungsten:
**Issue number:**
N/A

**Description of changes:**
This changes the build workflow and tools workflow so that instead of using self-hosted runners, we use Github-hosted 16-core runners.

**Testing done:**
See this PR's actions workflow.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,10 +15,12 @@ on:
     branches: [develop]
 jobs:
   build:
-    runs-on: [self-hosted, linux, x64]
+    runs-on:
+      group: bottlerocket
+      labels: bottlerocket_ubuntu-latest_16-core
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.61.0
+      - run: rustup default 1.61.0 && rustup component add rustfmt && rustup component add clippy
       - run: make build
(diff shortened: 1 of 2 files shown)


etungsten:
Push above adds `rustfmt` and `clippy` to the default rust toolchain.

etungsten:
So the build takes about 17 mins. If that's too long, we can use 32-core runners instead of 16 cores

stmcginnis:
> So the build takes about 17 mins. If that's too long, we can use 32-core runners instead of 16 cores

Do you have any metrics for the 32-core runners? If that takes us from 17 minutes to 7, that's one thing. But if it takes us from 17 to 15 minutes, then probably not really worth the extra expense.

stmcginnis (approved):
I think we should probably go ahead with this, then we can decide later whether 32-core runners are worth it.

jpculp (approved):

etungsten:
17 mins is actually on par with our current set-up surprisingly.

webern:
> 17 mins is actually on par with our current set-up surprisingly.

There's 4-5 minutes of integ tests that don't really speed up with more cores.

PR #704, opened 2022-12-13
controller: Resource creation waits for test

ecpullen:
Resources now wait for a test that requires them before they are created. This along with an addition to `is_delete_required()` that checks to make sure the resource exists before deleting it solve a race condition where a resource that was applied to a cluster slightly before its corresponding test would automatically be cleaned up before it was created.

**Issue number:**

N/A

**Description of changes:**

Resources now wait for a test that requires them before they are created. This along with an addition to `is_delete_required()` that checks to make sure the resource exists before deleting it solve a race condition where a resource that was applied to a cluster slightly before its corresponding test would automatically be cleaned up before it was created.

Adds a new wait action, WaitForDependent that signifies that no tests or resources require that resource.

**Testing done:**

Applied a cluster and instances yaml without the proposed changes and the instances resource was instantly cleaned up. After this pr, applying the same yaml, instance resource is not automatically deleted.

**Terms of contribution:**

### DIFF
--- a/controller/src/resource_controller/mod.rs
+++ b/controller/src/resource_controller/mod.rs
@@ -103,6 +103,9 @@ async fn do_creation_action(r: ResourceInterface, action: CreationAction) -> Res
                 conflict
             );
         }
+        CreationAction::WaitForDependent => {
+            debug!("'{}' is waiting for test that requires it", r.name());
+        }
         CreationAction::AddResourceFinalizer => {
             let _ = r
                 .resource_client()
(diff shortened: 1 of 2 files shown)


webern (approved):

controller/src/resource_controller/mod.rs
```diff
@@ -103,6 +103,9 @@ async fn do_creation_action(r: ResourceInterface, action: CreationAction) -> Res
                 conflict
             );
         }
+        CreationAction::WaitForDependent => {
+            debug!("'{}' is waiting for test that requires it", r.name(),);
```
stmcginnis: Looks like the last `,` isn't needed?

stmcginnis (approved):
Looks fine to me.

etungsten (approved):

mjsterckx (approved):

ecpullen:
^ remove extra ','

etungsten (approved):

PR #705, opened 2022-12-13
sonobuoy: Provided updates to status

ecpullen:
**Issue number:**

Related to #679 

**Description of changes:**

Sonobuoy now sends updated every 30s with the current results.

**Testing done:**

Tested with `cargo make test` and verified that the status was updated.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/sonobuoy.rs
+++ b/bottlerocket/agents/src/sonobuoy.rs
@@ -5,18 +5,35 @@ use model::{Outcome, TestResults};
 use serde_json::Value;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::path::Path;
 use std::process::Command;
 use std::time::Duration;
+use test_agent::InfoClient;
 
 /// Runs the sonobuoy conformance tests according to the provided configuration and returns a test
 /// result at the end.
-pub async fn run_sonobuoy(
+pub async fn run_sonobuoy<I>(
     kubeconfig_path: &str,
     e2e_repo_config_path: Option<&str>,
     sonobuoy_config: &SonobuoyConfig,
     results_dir: &Path,
-) -> Result<TestResults, error::Error> {
+    info_client: &I,
+) -> Result<TestResults, error::Error>
+where
+    I: InfoClient,
+{
+    let mut results = TestResults {
+        outcome: Outcome::InProgress,
+        other_info: Some("Starting test".to_string()),
+        ..Default::default()
+    };
+    info_client
+        .send_test_update(results.clone())
+        .await
+        .err()
+        .iter()
+        .for_each(|e| error!("Unable to send test update: {}", e));
     let kubeconfig_arg = vec!["--kubeconfig", kubeconfig_path];
     let version = sonobuoy_config
         .kubernetes_version
@@ -57,6 +74,13 @@ pub async fn run_sonobuoy(
     // TODO - log something or check what happened?
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
+    results.other_info = Some("Checking status".to_string());
+    info_client
+        .send_test_update(results)
+        .await
+        .err()
+        .iter()
+        .for_each(|e| error!("Unable to send test update: {}", e));
     info!("Sonobuoy testing has started, waiting for status to be available");
     tokio::time::timeout(
         Duration::from_secs(300),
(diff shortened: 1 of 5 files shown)


bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -5,18 +5,30 @@ use model::{Outcome, TestResults};
...
+) -> Result<TestResults, error::Error>
+where
+    I: InfoClient,
+{
+    let mut results = TestResults {
+        outcome: Outcome::InProgress,
+        other_info: Some("Starting test".to_string()),
+        ..Default::default()
+    };
+    info_client.send_test_update(results.clone()).await?;
```
webern: Do we want to continue or bomb when errors like this one occur?

```suggestion
    info_client.send_test_update(results.clone()).await.ok_or_else(|e| {
        error!("Unable to sent test update: {}", e);
    });
```

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -57,6 +69,8 @@ pub async fn run_sonobuoy(
     // TODO - log something or check what happened?
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
+    results.other_info = Some("Checking status".to_string());
+    info_client.send_test_update(results).await?;
```
webern: Same as previous comment. Do we want to bring down the house of cards if the Kubernetes API is unreachable for a millisecond?

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -181,6 +208,9 @@ pub async fn wait_for_sonobuoy_results(
         let run_status: serde_json::Value = match serde_json::from_str(&stdout) {
             Ok(run_status) => run_status,
             Err(err) => {
+                results.other_info = Some(format!("Status failed '{}'", retries));
+                info_client.send_test_update(results.clone()).await?;
```
webern: Same as previous comments, we are taking on additional risk of a transient network blip ruining the party.

webern (approved):
I'm concerned with the increase in opportunities for a momentary issue reaching the k8s API causing the program to exit. We should consider logging errors in areas where we might recover.

ecpullen:
^ Errors from the info client are printed to error instead of breaking.

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -5,18 +5,35 @@ use model::{Outcome, TestResults};
...
+        outcome: Outcome::InProgress,
+        other_info: Some("Starting test".to_string()),
+        ..Default::default()
+    };
+    info_client
+        .send_test_update(results.clone())
+        .await
+        .err()
+        .iter()
+        .for_each(|e| error!("Unable to sent test update: {}", e));
```
webern:
```suggestion
        .for_each(|e| error!("Unable to send test update: {}", e));
```

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -57,6 +74,13 @@ pub async fn run_sonobuoy(
     // TODO - log something or check what happened?
     ensure!(status.success(), error::SonobuoyRunSnafu);
 
+    results.other_info = Some("Checking status".to_string());
+    info_client
+        .send_test_update(results)
+        .await
+        .err()
+        .iter()
+        .for_each(|e| error!("Unable to sent test update: {}", e));
```
webern:
```suggestion
        .for_each(|e| error!("Unable to send test update: {}", e));
```

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -197,10 +242,20 @@ pub async fn wait_for_sonobuoy_results(
...
+            }
+        }
+
+        info!("Current status: {:?}", results);
+        info_client
+            .send_test_update(results.clone())
+            .await
+            .err()
+            .iter()
+            .for_each(|e| error!("Unable to sent test update: {}", e));
```
webern:
```suggestion
            .for_each(|e| error!("Unable to send test update: {}", e));
```

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -181,6 +218,14 @@ pub async fn wait_for_sonobuoy_results(
         let run_status: serde_json::Value = match serde_json::from_str(&stdout) {
             Ok(run_status) => run_status,
             Err(err) => {
+                results.other_info = Some(format!("Status failed '{}'", retries));
+                info_client
+                    .send_test_update(results.clone())
+                    .await
+                    .err()
+                    .iter()
+                    .for_each(|e| error!("Unable to sent test update: {}", e));
```
webern:
```suggestion
                    .for_each(|e| error!("Unable to send test update: {}", e));
```

webern (approved):

bottlerocket/agents/src/sonobuoy.rs
ecpullen: Fixed.

bottlerocket/agents/src/sonobuoy.rs
```diff
@@ -426,6 +527,18 @@ mod test_sonobuoy {
...
+    #[test]
+    fn test_process_incomplete_results_progress_object() {
+        let result =
+            process_incomplete_sonobuoy_test_results(
+                &json!({"plugins":[{"plugin":"e2e","progress":{"name":"e2e","node":"global","timestamp":"2022-12-08T15:37:23.007805243Z","msg":"Test Suite completed","total":1,"completed":1},"status":"complete","result-status":"timed-out","result-counts":{"failed":1}}]})).unwrap();
+        assert_eq!(result.num_passed, 1);
+        assert_eq!(result.num_failed, 0);
+        assert_eq!(result.num_skipped, 0);
+        assert_eq!(result.outcome, Outcome::InProgress);
+        assert_eq!(result.other_info.unwrap(), "Running");
```
etungsten: I'm slightly confused about this unit test. The json there says the status is complete but we're asserting the result is `Outcome::InProgess`.

stmcginnis (approved):
Looks good to me!

bottlerocket/agents/src/sonobuoy.rs
ecpullen: It's because it is using the incomplete status results.

etungsten: `"result-status":"timed-out"`
Is it because the `result-status` isn't complete? If we're going by the codepath, `wait_for_sonobuoy_results` returns when the `status` is no longer `running` so this case should theoretically never happen?

ecpullen: Correct that code path shouldn't happen, I can delete the test if you want.

etungsten: Yes, I think we should delete this unit test if we don't expect this to ever happen. The presence of this unit test makes me think this is a possibility we want to test for (a very weird one at that).

etungsten (approved):

PR #725, opened 2022-12-21
Add ECS Workload Agent

### LINKED ISSUE #617: workload test agent for ECS
Create a workload test agent, as described in #419, for running workload tests on ECS.

stmcginnis:
**Issue number:**

Closes: #617

**Description of changes:**

This adds a new test agent for running workload testing in an ECS cluster.

**Testing done:**

See details of testing [in comment below](https://github.com/bottlerocket-os/bottlerocket-test-system/pull/725#issuecomment-1375925243)

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
+++ b/bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
@@ -1,0 +1,582 @@
+/*!
+
+This is a test-agent for running workload tests on ECS.
+It expects to be run in a pod launched by the TestSys controller.
+
+To build the container for the workload test agent, run `make ecs-workload-agent` from the
+root directory of this repository.
+
+Here is an example manifest for deploying the test definition for the workload test agent:
+
+```yaml
+apiVersion: testsys.system/v1
+kind: Test
+metadata:
+  name: ecs-workload
+  namespace: testsys
+spec:
+  agent:
+    configuration:
+      region: "us-east-2"
+      clusterName: "bottlerocket"
+      assumeRole: "arn:aws:sts::999807767999:assumed-role/InstanceAdmin/i-89898f9c69b228989"
+      tests:
+      - name: nvidia-workload
+        image: testsys-nvidia-workload-test:v0.0.3
+        gpu: true
+      - name: webserver
+        image: nginx:latest
+    image: <your ecs-workload-agent image URI>
+    name: ecs-workload-test-agent
+    keepRunning: true
+    timeout: "5000"
+  resources: []
+```
+
+!*/
+
+use agent_utils::aws::aws_config;
+use agent_utils::init_agent_logger;
+use async_trait::async_trait;
+use aws_sdk_ec2::types::SdkError;
+use aws_sdk_ecs::error::DescribeTaskDefinitionErrorKind;
+use aws_sdk_ecs::model::{
+    Compatibility, ContainerDefinition, LaunchType, ResourceRequirement, ResourceType, SortOrder,
+    TaskDefinition, TaskStopCode,
+};
+use bottlerocket_agents::error::{self, Error};
+use bottlerocket_types::agent_config::{
+    EcsWorkloadTestConfig, WorkloadTest, AWS_CREDENTIALS_SECRET_NAME,
+};
+use log::info;
+use model::{Outcome, SecretName, TestResults};
+use snafu::{OptionExt, ResultExt};
+use std::time::Duration;
+use test_agent::{
+    BootstrapData, ClientError, DefaultClient, DefaultInfoClient, InfoClient, Runner, Spec,
+    TestAgent,
+};
+
+pub const DEFAULT_TASK_DEFINITION_PREFIX: &str = "testsys-bottlerocket-";
+
+struct EcsWorkloadTestRunner {
+    config: EcsWorkloadTestConfig,
+    aws_secret_name: Option<SecretName>,
+}
+
+#[async_trait]
+impl<I> Runner<I> for EcsWorkloadTestRunner
+where
+    I: InfoClient,
+{
+    type C = EcsWorkloadTestConfig;
+    type E = Error;
+
+    /// Initialize a new instance of our workload test runner.
+    async fn new(spec: Spec<Self::C>, _info_client: &I) -> Result<Self, Self::E> {
+        info!("Initializing ECS workload test agent...");
+        Ok(Self {
+            config: spec.configuration,
+            aws_secret_name: spec.secrets.get(AWS_CREDENTIALS_SECRET_NAME).cloned(),
+        })
+    }
+
+    /// Run the set of configured workload tests and return the results.
+    async fn run(&mut self, info_client: &I) -> Result<TestResults, Self::E> {
+        let config = aws_config(
+            &self.aws_secret_name.as_ref(),
+            &self.config.assume_role,
+            &None,
+            &self.config.region,
+            false,
+        )
+        .await?;
+        let ecs_client = aws_sdk_ecs::Client::new(&config);
+
+        info!("Waiting for registered container instances...");
+
+        tokio::time::timeout(
+            Duration::from_secs(120),
+            wait_for_cluster_ready(&ecs_client, &self.config.cluster_name),
+        )
+        .await
+        .context(error::InstanceTimeoutSnafu)??;
+
+        // First loop through and make sure we have task definitions for each of our tests.
+        // We don't want to starts tasks if one of the defined tets is not able to be run.
+        let mut task_def_arns = Vec::new();
+        for test_def in &self.config.tests {
+            let task_def_arn = create_or_find_task_definition(&ecs_client, test_def).await?;
+
+            task_def_arns.push(task_def_arn);
+        }
+
+        let mut results = TestResults {
+            outcome: Outcome::InProgress,
+            other_info: Some("Starting test".to_string()),
+            ..Default::default()
+        };
+        info_client
+            .send_test_update(results.clone())
+            .await
+            .err()
+            .iter()
+            .for_each(|e| log::error!("Unable to send test update: {}", e));
+
+        // Now that we have the full set of task definitions to run, loop through and kick them all off.
+        let mut task_arns = Vec::new();
+        for task_def_arn in task_def_arns {
+            info!("Running task '{}'", task_def_arn);
+            let run_task_output = ecs_client
+                .run_task()
+                .cluster(&self.config.cluster_name)
+                .task_definition(task_def_arn)
+                .launch_type(LaunchType::Ec2)
+                .send()
+                .await
+                .context(error::TaskRunCreationSnafu)?;
+            let run_task_arns: Vec<String> = run_task_output
+                .tasks()
+                .map(|tasks| {
+                    tasks
+                        .iter()
+                        .filter_map(|task| task.task_arn().map(|arn| arn.to_string()))
+                        .collect()
+                })
+                .unwrap();
+            for task_arn in &run_task_arns {
+                task_arns.push(task_arn.clone());
+            }
+        }
+
+        info!("Waiting for tasks to complete...");
+        results.other_info = Some("Checking status".to_string());
+        info_client
+            .send_test_update(results)
+            .await
+            .err()
+            .iter()
+            .for_each(|e| log::error!("Unable to send test update: {}", e));
+
+        // Some workload tests could take awhile to complete. This is a long timeout, but we need to make sure there is
+        // enough time for any larger tests to finish.
+        // Note: may want to make this configurable by test definition yaml so it can be customized by the actual tests
+        // being run.
+        match tokio::time::timeout(
+            Duration::from_secs(1800),
+            wait_for_test_completion(
+                &ecs_client,
+                &self.config.cluster_name,
+                &task_arns,
+                info_client,
+            ),
+        )
+        .await
+        {
+            Ok(results) => results,
+            Err(_) => {
+                let mut res =
+                    test_results(&ecs_client, &self.config.cluster_name, &task_arns).await;
+
+                // We've timed out, but there is a window where our final check could actually report completion.
+                // To guard against this, check if still in progress and mark failed only if so.
+                if let Ok(mut result) = res {
+                    if result.outcome == Outcome::InProgress {
+                        result.outcome = Outcome::Timeout;
+                    }
+                    res = Ok(result);
+                }
+
+                res
+            }
+        }
+    }
+
+    /// Terminate the test runs. This is a noop for this agent.
+    async fn terminate(&mut self) -> Result<(), Self::E> {
+        Ok(())
+    }
+}
+
+/// Loops waiting for the workload test tasks to complete.
+async fn wait_for_test_completion<I>(
+    ecs_client: &aws_sdk_ecs::Client,
+    cluster_name: &str,
+    task_arns: &[String],
+    info_client: &I,
+) -> Result<TestResults, Error>
+where
+    I: InfoClient,
+{
+    let mut retries: u64 = 0;
+    loop {
+        let mut results = test_results(ecs_client, cluster_name, task_arns).await?;
+        // If tests have all passed or failed, return. Otherwise still in progress so wait and retry.
+        if results.outcome == Outcome::Pass || results.outcome == Outcome::Fail {
+            info_client
+                .send_test_update(results.clone())
+                .await
+                .err()
+                .iter()
+                .for_each(|e| log::error!("Unable to send test update: {}", e));
+            return Ok(results);
+        }
+
+        retries += 1;
+        results.other_info = Some(format!("Waiting for test completion: attempt {}", retries));
+        info_client
+            .send_test_update(results.clone())
+            .await
+            .err()
+            .iter()
+            .for_each(|e| log::error!("Unable to send test update: {}", e));
+        tokio::time::sleep(Duration::from_secs(2)).await;
+    }
+}
+
+/// Gets the status from our test tasks and populates a `TestResults` struct with the outcome.
+async fn test_results(
+    ecs_client: &aws_sdk_ecs::Client,
+    cluster_name: &str,
+    task_arns: &[String],
+) -> Result<TestResults, Error> {
+    let tasks = ecs_client
+        .describe_tasks()
+        .cluster(cluster_name)
+        .set_tasks(Some(task_arns.to_vec()))
+        .send()
+        .await
+        .context(error::TaskDescribeSnafu)?
+        .tasks()
+        .map(|tasks| tasks.to_owned())
+        .context(error::NoTaskSnafu)?;
+    let passed_count = tasks
+        .iter()
+        .filter(|task| task.last_status() == Some("STOPPED"))
+        .filter(|task| task.stop_code() == Some(&TaskStopCode::EssentialContainerExited))
+        .filter(|task| {
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() != Some(0))
+                .count()
+                == 0
+        })
+        .count() as i32;
+    let failed_count = tasks
+        .iter()
+        .filter(|task| task.last_status() == Some("STOPPED"))
+        .filter(|task| task.stop_code() == Some(&TaskStopCode::EssentialContainerExited))
+        .filter(|task| {
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() != Some(0))
+                .count()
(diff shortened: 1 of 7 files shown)


stmcginnis:
Note to self: always, always, always rebuild after rebasing.

stmcginnis:
Still testing to be done, but marking this as ready for review since I think the error is something in my local test setup and not necessarily anything with the agent itself.

bottlerocket/types/src/agent_config.rs
```diff
@@ -495,12 +495,25 @@ fn k8s_version_valid() {
...
+    pub gpu: bool,
 }
 
 #[derive(Debug, Clone, Default, Serialize, Deserialize, Configuration, Builder)]
 #[serde(rename_all = "camelCase")]
 #[crd("Test")]
 pub struct WorkloadConfig {
-    pub kubeconfig_base64: String,
-    pub plugins: Vec<WorkloadTest>,
+    pub kubeconfig_base64: Option<String>,
```
ecpullen: Is there a reason this was changed to an option?

bottlerocket/types/src/agent_config.rs
```diff
@@ -495,12 +495,25 @@ fn k8s_version_valid() {
...
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize, Configuration, Builder)]
+#[serde(rename_all = "camelCase")]
+#[crd("Test")]
+pub struct EcsWorkloadTestConfig {
+    pub region: Option<String>,
+    pub cluster_name: String,
+    /// The role that should be assumed for this test agent.
+    pub assume_role: Option<String>,
```
ecpullen, line 517: We probably need an assumed role for the k8s workload too.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+    }
+
+    /// Terminate the test runs. This is a noop for this agent.
+    async fn terminate(&mut self) -> Result<(), Self::E> {
+        Ok(())
+    }
+}
+
+/// Loops waiting for the workload test tasks to complete.
```
ecpullen, line 201: Should this send updates to the crd like sonobuoy does now?

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,499 @@
...
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() != Some(0))
+                .count()
+                == 0
+        })
+        .count() as i32;
+    Ok(TestResults {
+        outcome: if 1 == running_count {
```
ecpullen: Why is this `1 == running_count`

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,499 @@
...
+            Outcome::Pass
+        } else {
+            Outcome::Fail
+        },
+        num_passed: running_count as u64,
+        num_failed: (1 - running_count) as u64,
```
ecpullen: I'm unsure what `running_count` represents. Is it either `1` or `0` or is the number of running tasks?

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,499 @@
...
+            .clusters(cluster)
+            .send()
+            .await
+            .context(error::ClusterDescribeSnafu)?
+            .clusters()
+            .context(error::NoTaskSnafu)?
+            .first()
+            .context(error::NoTaskSnafu)?
+            .clone();
+
```
ecpullen, line 315: Can we add `info!("Waiting for cluster to have registered instances")`

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
```diff
@@ -68,8 +68,12 @@ where
 
     async fn run(&mut self, info_client: &I) -> Result<TestResults, Self::E> {
         info!("Decoding kubeconfig for test cluster");
-        base64_decode_write_file(&self.config.kubeconfig_base64, TEST_CLUSTER_KUBECONFIG_PATH)
-            .await?;
+        let kubeconfig_base64 = match &self.config.kubeconfig_base64 {
+            Some(kc) => kc.to_string(),
+            None => "".to_string(),
+        };
```
ecpullen: Should `kubeconfig_base64` be an option? If so this may be a bit cleaner.

```suggestion
        let kubeconfig_base64 = &self.config.kubeconfig_base64.to_owned().unwrap_or_default()
```

ecpullen (approved):

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,499 @@
...
+            &err.kind,
+            DescribeTaskDefinitionErrorKind::ClientException(_)
+        ) {
+            return create_task_definition(ecs_client, &task_def_name, test_def).await;
+        }
+    }
+
+    // There is a task definition with the right name, now we need to find the right
+    // revision that matches our image:tag.
+    if let task_arn @ Ok(_) = find_task_rev(ecs_client, &task_def_name, test_def).await {
```
ecpullen, line 353:
```suggestion
    if let Some(task_arn) = find_task_rev(ecs_client, &task_def_name, test_def).await.ok() {
```

bottlerocket/types/src/agent_config.rs
stmcginnis: I'm trying to reuse this struct for ECS testing. With K8s, this is needed for Sonobuoy to have the kubeconfig to speak to the cluster. With ECS there is no such thing. So I switched it to `Option` to allow either case.

ecpullen: So the `EcsWorkloadTestConfig` is not going to be used?

bottlerocket/types/src/agent_config.rs
stmcginnis: Let's leave that for a separate PR. With the SDK client used in the ECS testing it's an easy matter of just passing this in to the client initialization. I think the EKS/K8s code will need a little more work to enable that.

stmcginnis: Filed https://github.com/bottlerocket-os/bottlerocket-test-system/issues/732 so we don't lose that.

bottlerocket/types/src/agent_config.rs
stmcginnis: Oops! Now that I actually look again, I think I was originally going down the route of trying to use `WorkloadConfig` for both types of tests, then realized I needed other things that required the `EcsWorkloadTestConfig` struct. I'll change this back.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: Another mistake as I was working through things. This should actually be the length of `task_arns`.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: Can you point me to where things are updated in the latest? I'm missing that part. I don't see what changed in that respect in https://github.com/bottlerocket-os/bottlerocket-test-system/pull/690/files#diff-bb5e446ed77b2db083260a5cd9e88200daf3f723eec7d88d9ba1840f7b98fd8d

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: It would be the running tasks. In the existing `ecs-test-agent` this number is passed in as part of the config, but it should just be the number of total tasks.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: Good call - added!

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: This isn't correct. We want to get the `Result<String, Error>`, not just the `String`.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
ecpullen: https://github.com/bottlerocket-os/bottlerocket-test-system/blob/1e8122cb99c4b45bc2b58730bb1b50d9d6da1369/bottlerocket/agents/src/sonobuoy.rs#L26-L36

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
ecpullen: Isn't this only capturing the case where `find_task_rev` is `Ok(_)`. We are throwing the error away anyways. Could also just do  ` if let Ok(task_arn) = find_task_rev(ecs_client, &task_def_name, test_def).await {`.

stmcginnis: We're not trying to capture just `task_arn` though. We need to return a `Result<String, Error>`, so if we captured just the String on success we would just have to recreate the `Result<String, Error>` again. With this, if it is successful we just take that `Result` and return it. If it is _not_ successful, then we continue on and make the call to `create_task_definition`.

So we can change it that way, but it wouldn't be very efficient to just pull out the String and put it back into a new `Result`.

ecpullen (approved):

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: I think I've got it. Please take a look at the [diff from the last push](https://github.com/bottlerocket-os/bottlerocket-test-system/compare/1ca83cb85952a2b6c46b24a5426d15d3fbb62913..8dadb103076282231e9d7896b6fff0ccf8c19749) and let me know if that is not right.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,545 @@
...
+                &self.config.cluster_name,
+                &task_arns,
+                info_client,
+            ),
+        )
+        .await
+        {
+            Ok(results) => results,
+            Err(_) => test_results(&ecs_client, &self.config.cluster_name, &task_arns).await,
+        }
```
ecpullen: Does this mean the task only has 30s to run to completion?

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,545 @@
...
+            .await
+            .err()
+            .iter()
+            .for_each(|e| log::error!("Unable to send test update: {}", e));
+        tokio::time::sleep(Duration::from_secs(2)).await;
+    }
+}
+
+/// Gets the status from our test tasks and populates a `TestResults` struct with the outcome.
+async fn test_results(
```
ecpullen, line 238: I think we need a different way of return an incomplete test than `Outcome::Fail`. If a task immediately fails, it will still loop until the timeout I commented on above.

ecpullen: Maybe if some are still running use `Outcome::InProgress`

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
ecpullen: Looks good.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: Hmm, yeah, that's way too short for these. Good catch.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
stmcginnis: I was wondering if there was a reason for defaulting to `Fail`.

https://github.com/bottlerocket-os/bottlerocket-test-system/blob/develop/bottlerocket/agents/src/bin/ecs-test-agent/main.rs#L169-L173

Maybe to have the result be failure if the test execution times out?

https://github.com/bottlerocket-os/bottlerocket-test-system/blob/develop/bottlerocket/agents/src/bin/ecs-test-agent/main.rs#L107-L114

But I think it makes more sense to report that it is in progress than to indicate a failure when there is none.

stmcginnis: Added check to make sure we properly report failure on timeout.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,563 @@
...
+            Ok(results) => results,
+            Err(_) => {
+                let mut res =
+                    test_results(&ecs_client, &self.config.cluster_name, &task_arns).await;
+
+                // We've timed out, but there is a window where our final check could actually report completion. To
+                // guard against this, check if still in progress and marked failed only if so.
+                if let Ok(mut result) = res {
+                    if result.outcome == Outcome::InProgress {
+                        result.outcome = Outcome::Fail;
```
ecpullen: nit: Outcome::Timeout is probably more accurate.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,563 @@
...
+        .context(error::NoTaskSnafu)?;
+    let running_count = tasks
+        .iter()
+        .filter(|task| task.last_status() == Some("STOPPED"))
+        .filter(|task| task.stop_code() == Some(&TaskStopCode::EssentialContainerExited))
+        .filter(|task| {
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() != Some(0))
```
ecpullen, line 261: Should we count on failed tasks so we can accurately report them in test results? I don't think we should mark a test as a failure if it hasn't finished executing.

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,563 @@
...
+                == 0
+        })
+        .count() as i32;
+    let task_count = task_arns.len() as i32;
+    Ok(TestResults {
+        outcome: if task_count == running_count {
+            Outcome::Pass
+        } else {
+            Outcome::InProgress
+        },
```
ecpullen:
```suggestion
        outcome: if task_count == running_count + fail_count {
        if fail_count > 0 || pass_count == 0{
            Outcome::Fail
           } else {
           Outcome::Pass
        } else {
            Outcome::InProgress
        },
```

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
ecpullen: In this case, `num_passed + num_failed < task_count` unless all tests completed.

ecpullen: We should report failed tasks in the status updates, but wait until all tasks finish before reporting results.

stmcginnis: I like that better. 👍

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,582 @@
...
+        .filter(|task| task.last_status() == Some("STOPPED"))
+        .filter(|task| task.stop_code() == Some(&TaskStopCode::EssentialContainerExited))
+        .filter(|task| {
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() == Some(0))
+                .count()
+                == 0
+        })
```
ecpullen, line 277:
```suggestion
                .filter(|container| container.exit_code() != Some(0))
                .count()
                != 0
        })
```

bottlerocket/agents/src/bin/ecs-workload-agent/main.rs
```diff
@@ -0,0 +1,582 @@
...
+        .iter()
+        .filter(|task| task.last_status() == Some("STOPPED"))
+        .filter(|task| task.stop_code() == Some(&TaskStopCode::EssentialContainerExited))
+        .filter(|task| {
+            task.containers()
+                .unwrap_or_default()
+                .iter()
+                .filter(|container| container.exit_code() == Some(0))
+                .count()
+                == 0
```
ecpullen: Should it error if any containers exit non-zero instead of requiring all of them?
```suggestion
                .filter(|container| container.exit_code() != Some(0))
                .count()
                != 0
```

stmcginnis: Yeah, still getting used to these long filter calls and was breaking my brain. I see what you're saying now - will update.

ecpullen (approved):
LGTM. Can you verify that the nvidia test works with this agent?

stmcginnis:
> LGTM. Can you verify that the nvidia test works with this agent?

Able to deploy an ECS cluster instance running Bottlerocket 1.11.1, `aws-ecs-1-nvidia`, using the following:

```yaml
apiVersion: testsys.system/v1
kind: Test
metadata:             
  name: ecs-workload              
  namespace: testsys 
spec:             
  agent:                 
    configuration:    
      region: "us-east-2"
      clusterName: "brtest"                                                    
      tests:                          
      - name: nvidia-workload
        image: public.ecr.aws/p0f9g6e8/gpu-tests:latest
        gpu: true                   
    image: ecs-workload-agent:demo
    name: ecs-workload-test-agent
    keepRunning: true
    timeout: "5000"
    secrets:
      awsCredentials: "aws-creds"
  resources:
  - brtest-instances
  - brtest
  dependsOn: []
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: brtest
  namespace: testsys
spec:
  agent:
    name: ecs-provider
    image: ecs-resource-agent:demo
    keepRunning: true
    configuration:
      clusterName: brtest
      region: "us-east-2"
    secrets:
      awsCredentials: "aws-creds"
  dependsOn: []
  destructionPolicy: never
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: brtest-instances
  namespace: testsys
spec:
  agent:
    name: ec2-provider
    image: ec2-resource-agent:demo
    keepRunning: true
    configuration:
      clusterName: brtest
      clusterType: ecs
      instanceCount: 2
      instanceProfileArn: ${brtest.iamInstanceProfileArn}
      nodeAmi: "ami-0ba78cb33c07e2333"
      region: "us-east-2"
      subnetIds: ${brtest.publicSubnetIds}
      instanceTypes: ["g4dn.xlarge"]
    secrets:
      awsCredentials: "aws-creds"
  dependsOn: [brtest]
  destructionPolicy: never
```

Able to create cluster, launch and join EC2 instances, and execute NVIDIA workload test. Logs from the workload test:

```bash
$ k -n testsys logs pod/ecs-workload-qzxnr
[2023-01-09T16:27:02Z INFO  ecs_workload_agent] Initializing ECS workload test agent...
[2023-01-09T16:27:02Z INFO  ecs_workload_agent] Waiting for registered container instances...
[2023-01-09T16:27:02Z INFO  ecs_workload_agent] Waiting for cluster to have registered instances
[2023-01-09T16:27:04Z INFO  ecs_workload_agent] Waiting for cluster to have registered instances
[2023-01-09T16:27:06Z INFO  ecs_workload_agent] Waiting for cluster to have registered instances
[2023-01-09T16:27:08Z INFO  ecs_workload_agent] Waiting for cluster to have registered instances
[2023-01-09T16:27:09Z INFO  ecs_workload_agent] Running task 'arn:aws:ecs:us-east-2:861807767978:task-definition/testsys-bottlerocket-nvidia-workload:1'
[2023-01-09T16:27:09Z INFO  ecs_workload_agent] Waiting for tasks to complete...
[2023-01-09T16:27:57Z INFO  test_agent::agent] Test execution finished without returning an error.
[2023-01-09T16:27:57Z INFO  test_agent::agent] Test output tarball created.
[2023-01-09T16:27:57Z INFO  test_agent::agent] 'keep_running' is true.
```

Stopped task in the ECS cluster shows exited with exit code 0.

etungsten (approved):

PR #731, opened 2023-01-04
workload: Remove `canonicalize` from plugin file path

ecpullen:
**Issue number:**

N/A

**Description of changes:**

There was a bug in the workload agent where the plugin's file path couldn't be canonicalized.

**Testing done:**

Tested with `cargo make test`.
```bash
 NAME                         TYPE      STATE      PASSED      SKIPPED      FAILED     STATUS    LAST UPDATE                BUILD ID
 x86-64-aws-k8s-124-test      Test      pass       1           0            0                     2023-01-04T14:58:58Z      3ebb33a9
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/error.rs
+++ b/bottlerocket/agents/src/error.rs
@@ -106,12 +106,6 @@ pub enum Error {
         path: String,
     },
 
-    #[snafu(display("Invalid path: {path}"))]
-    BadPath {
-        source: std::io::Error,
-        path: String,
-    },
-
     #[snafu(display("Results location is invalid"))]
     ResultsLocation,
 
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Learned `canonicalize` will [fail if the path does not exist](https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors), which it won't at this point since we are just getting the path to use to write out the file contents later.

I can only assume when I tested this I must have had a file already in my path, but that won't be a possibility when running containerized.

This looks good. There really shouldn't be a need to get the canonical path anyway.

etungsten (approved):

PR #750, opened 2023-01-23
samples: added workload agent files

mjsterckx:
**Issue number:**

N/A

**Description of changes:**

Added sample files for ecs and k8s workload test agents. Updated `bottlerocket/samples/RUNBOOK.md` and `bottlerocket/types/agent_config.rs` with new guides and unit tests.

**Testing done:**

- New unit tests pass.

- ecs-workload-tests: nvidia-smoke-test passes
```
NAME                                      TYPE             STATE             PASSED         SKIPPED        FAILED
x86-64-aws-ecs-1-nvidia                   Resource         completed
x86-64-aws-ecs-1-nvidia-instances         Resource         completed
x86-64-aws-ecs-1-nvidia-test              Test             pass              1              0              0
```
- k8s-workload-tests: nvidia-smoke-test passes
```
NAME                                        TYPE             STATE            PASSED        SKIPPED        FAILED
x86-64-aws-k8s-124-nvidia                   Resource         completed
x86-64-aws-k8s-124-nvidia-instances         Resource         completed
x86-64-aws-k8s-124-nvidia-test              Test             pass             11            0              0
```
**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
+++ b/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
@@ -6,7 +6,7 @@ It expects to be run in a pod launched by the TestSys controller.
 You can configure the workload agent to run different types of plugins and tests.
 See `WorkloadConfig` for the different configuration values.
 
-To build the container for the workload test agent, run `make k8s-workload-agent-image` from the
+To build the container for the workload test agent, run `make k8s-workload-agent` from the
 root directory of this repository.
 
 Here is an example manifest for deploying the test definition for the workload test agent to a K8s cluster:
@@ -21,9 +21,10 @@ spec:
   agent:
     configuration:
       kubeconfigBase64: <Base64 encoded kubeconfig for the test cluster workload runs the tests in>
-      plugins:
+      tests:
       - name: nvidia-workload
         image: testsys-nvidia-workload-test:v0.0.3
+        gpu: false
     image: <your k8s-workload-agent image URI>
     name: workload-test-agent
     keepRunning: true
--- a/bottlerocket/samples/RUNBOOK.md
+++ b/bottlerocket/samples/RUNBOOK.md
@@ -69,6 +69,37 @@ EOF
 " 2> /dev/null
 ```
 
+### Workload Testing on `aws-ecs` Variants
+
+_Note_: An example of a workload test that can be used with the `ecs-workload-agent` is the `nvidia-smoke` test. See the `nvidia-smoke` [README.md](../tests/workload/nvidia-smoke/README.md) for instructions on how to build and use its image.
+
+```bash
+CLUSTER_NAME="x86-64-aws-ecs-1-nvidia"
+OUTPUT_FILE="${CLUSTER_NAME}.yaml"
+VARIANT="aws-ecs-1-nvidia"
+ARCHITECTURE="x86_64"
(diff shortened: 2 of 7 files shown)


webern (commented):
This looks great provided we stick with the original pattern. However, I have some concerns about the RUNBOOK becoming incorrect when changes are made to the yaml files. And more generally, that the RUNBOOK is an example of "code duplication".

We could document the purpose of the yaml file within the yaml file itself, and to document the environment variables with comments at their sites in the YAML files. Then the Runbook could give a general example of how to do the substitution and give the code. @etungsten @ecpullen 

The first principal that I'm trying to solve for is that any code not tested in CI will definitely be broken at some point in the future. And this Markdown code fits that description.

ecpullen:
> This looks great provided we stick with the original pattern. However, I have some concerns about the RUNBOOK becoming incorrect when changes are made to the yaml files. And more generally, that the RUNBOOK is an example of "code duplication".

The next step is creating a Makefile to reduce the length of the runbook.

> We could document the purpose of the yaml file within the yaml file itself, and to document the environment variables with comments at their sites in the YAML files. Then the Runbook could give a general example of how to do the substitution and give the code. @etungsten @ecpullen
>
> The first principal that I'm trying to solve for is that any code not tested in CI will definitely be broken at some point in the future. And this Markdown code fits that description.

The makefile targets are the ones that will be used for release testing, so they will not be able to become stale.

webern (approved):
> The makefile targets are the ones that will be used for release testing, so they will not be able to become stale.

I'm not enthused about the answer "we'll find out they are broken when we try to do a release", but I also don't want to block progress.

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
```diff
@@ -6,7 +6,7 @@ It expects to be run in a pod launched by the TestSys controller.
 You can configure the workload agent to run different types of plugins and tests.
 See `WorkloadConfig` for the different configuration values.
 
-To build the container for the workload test agent, run `make k8s-workload-agent-image` from the
+To build the container for the workload test agent, run `make k8s-workload-agent` from the
```
stmcginnis, line 9: Thanks for the clean up!

bottlerocket/samples/RUNBOOK.md
```diff
@@ -69,6 +69,37 @@ EOF
 " 2> /dev/null
 ```
 
+### Workload Testing on `aws-ecs` Variants
+
+_Note_: An example of a workload test that can be used with the `ecs-workload-agent` is the `nvidia-smoke` test. See [README.md](../tests/workload/nvidia-smoke/README.md) for instructions on how to build and use its image.
```
stmcginnis: Not something I think needs to hold anything up, but if you do end up updating for some other reason I think it would be good to change the link from just being "README.md" to something like "the nvidia-smoke README".

stmcginnis (approved):
Everything looks right to me. One suggestion noted, but again, not something I think should block this.

bottlerocket/samples/RUNBOOK.md
mjsterckx: I realized that that might get confusing but wasn't sure if that warranted more specification. Since you brought it up, might as well change it. Thanks!

ecpullen (approved):

PR #751, opened 2023-01-26
samples: added Makefile.toml to simplify runbook

mjsterckx:
**Issue number:**

N/A

**Description of changes:**

Added a Makefile to `bottlerocket/samples` that contains targets to create and run test yaml files. It calls the `create_sample_file.sh` script located in the `scripts` folder which generates the correct file based on arguments passed from the Makefile, using environment variables to populate the sample file (or default values if the variable is unset).

**Testing done:**

Created each sample with default values as well as overriding variables. Ran created samples on EKS and kind clusters.
For example:

- `cargo make create ecs-test` produces `x86-64-aws-ecs-1.yaml`:
```
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: x86-64-aws-ecs-1-test
  namespace: testsys
spec:
  agent:
    name: ecs-test-agent
    image: public.ecr.aws/bottlerocket-test-system/ecs-test-agent:v0.0.5
    keepRunning: true
    configuration:
      clusterName: ${x86-64-aws-ecs-1.clusterName}
      region: ${x86-64-aws-ecs-1.region}
      assumeRole: ~
  dependsOn: []
  resources: [x86-64-aws-ecs-1-instances, x86-64-aws-ecs-1]
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: x86-64-aws-ecs-1
  namespace: testsys
spec:
  agent:
    name: ecs-provider
    image: public.ecr.aws/bottlerocket-test-system/ecs-resource-agent:v0.0.5
    keepRunning: true
    configuration:
      clusterName: x86-64-aws-ecs-1
      region: us-west-2
      assumeRole: ~
  dependsOn: []
  destructionPolicy: onDeletion
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: x86-64-aws-ecs-1-instances
  namespace: testsys
spec:
  agent:
    name: ec2-provider
    image: public.ecr.aws/bottlerocket-test-system/ec2-resource-agent:v0.0.5
    keepRunning: true
    configuration:
      clusterName: ${x86-64-aws-ecs-1.clusterName}
      clusterType: ecs
      instanceCount: 2
      nodeAmi: ami-08cf4a2f9255f1cfb
      region: us-west-2
      instanceProfileArn: ${x86-64-aws-ecs-1.iamInstanceProfileArn}
      subnetIds: ${x86-64-aws-ecs-1.publicSubnetIds}
      instanceTypes: ["m5.large"]
      assumeRole: ~
  dependsOn: [x86-64-aws-ecs-1]
  destructionPolicy: onDeletion
```
- The following sequence of commands:
```
$ export CLUSTER_NAME=test-cluster
$ export ARCHITECTURE="aarch64"
$ export AGENT_IMAGE_VERSION=0.0.4
$ cargo make create sonobuoy-migration-test
```
produces `test-cluster-migration.yaml`:
```
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: test-cluster-test-1-initial
  namespace: testsys
spec:
  agent:
    name: sonobuoy-test-agent
    image: public.ecr.aws/bottlerocket-test-system/sonobuoy-test-agent:v0.0.4
    keepRunning: true
    configuration:
      region: ${test-cluster.region}
      assumeRole: ~
      kubeconfigBase64: ${test-cluster.encodedKubeconfig}
      plugin: "e2e"
      mode: "quick"
  dependsOn: []
  resources: [test-cluster-instances, test-cluster]
---  
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: test-cluster-test-2-migrate
  namespace: testsys
spec:
  agent:
    configuration:
      awsRegion: us-west-2
      instanceIds: ${test-cluster-instances.ids}
      migrateToVersion: v1.11.1
      tufRepo:
        metadataUrl: https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.24/aarch64
        targetsUrl: https://updates.bottlerocket.aws/targets
      assumeRole: ~
    image: public.ecr.aws/bottlerocket-test-system/migration-test-agent:v0.0.4
    name: migration-test-agent
    keepRunning: true
  resources: [test-cluster-instances, test-cluster]
  dependsOn: [test-cluster-test-1-initial]
---
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: test-cluster-test-3-migrated
  namespace: testsys
spec:
  agent:
    name: sonobuoy-test-agent
    image: public.ecr.aws/bottlerocket-test-system/sonobuoy-test-agent:v0.0.4
    keepRunning: true
    configuration:
      region: ${test-cluster.region}
      assumeRole: ~
      kubeconfigBase64: ${test-cluster.encodedKubeconfig}
      plugin: "e2e"
      mode: "quick"
  dependsOn: [test-cluster-test-2-migrate]
  resources: [test-cluster-instances, test-cluster]
---
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: test-cluster-test-4-migrate
  namespace: testsys
spec:
  agent:
    configuration:
      awsRegion: us-west-2
      instanceIds: ${test-cluster-instances.ids}
      migrateToVersion: v1.11.0
      tufRepo:
        metadataUrl: https://updates.bottlerocket.aws/2020-07-07/aws-k8s-1.24/aarch64
        targetsUrl: https://updates.bottlerocket.aws/targets
      assumeRole: ~
    image: public.ecr.aws/bottlerocket-test-system/migration-test-agent:v0.0.4
    name: migration-test-agent
    keepRunning: true
  resources: [test-cluster-instances, test-cluster]
  dependsOn: [test-cluster-test-3-migrated]
---
apiVersion: testsys.system/v1
kind: Test
metadata:
  name: test-cluster-test-5-final
  namespace: testsys
spec:
  agent:
    name: sonobuoy-test-agent
    image: public.ecr.aws/bottlerocket-test-system/sonobuoy-test-agent:v0.0.4
    keepRunning: true
    configuration:
      region: ${test-cluster.region}
      assumeRole: ~
      kubeconfigBase64: ${test-cluster.encodedKubeconfig}
      plugin: "e2e"
      mode: "quick"
  dependsOn: [test-cluster-test-4-migrate]
  resources: [test-cluster-instances, test-cluster]
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: test-cluster
  namespace: testsys
spec:
  agent:
    name: eks-provider
    image: public.ecr.aws/bottlerocket-test-system/eks-resource-agent:v0.0.4
    keepRunning: true
    configuration:
      creationPolicy: ifNotExists
      cluster_name: test-cluster
      region: us-west-2
      assumeRole: ~
  dependsOn: []
  destructionPolicy: onDeletion
---
apiVersion: testsys.system/v1
kind: Resource
metadata:
  name: test-cluster-instances
  namespace: testsys
spec:
  agent:
    name: ec2-provider
    image: public.ecr.aws/bottlerocket-test-system/ec2-resource-agent:v0.0.4
    keepRunning: true
    configuration:
      clusterName: ${test-cluster.clusterName}
      clusterType: eks
      instanceCount: 2
      instanceProfileArn: ${test-cluster.iamInstanceProfileArn}
      nodeAmi: ami-018b35ddb7c488284
      region: us-west-2
      subnetIds: ${test-cluster.publicSubnetIds}
      instanceTypes: ["m5.large"]
      assumeRole: ~
      endpoint: ${test-cluster.endpoint}
      certificate: ${test-cluster.certificate}
      clusterDnsIp: ${test-cluster.clusterDnsIp}
      securityGroups: ${test-cluster.securityGroups}
  dependsOn: [test-cluster]
  destructionPolicy: onDeletion
```

**Terms of contribution:**


bottlerocket/samples/RUNBOOK.md
```diff
@@ -191,6 +193,8 @@ EOF
 
 This codeblock assumes that your vSphere config file has been sourced. Specifically, the variables `GOVC_USERNAME`, `GOVC_PASSWORD`, `GOVC_DATACENTER`, `GOVC_DATASTORE`, `GOVC_URL`, `GOVC_NETWORK`, `GOVC_RESOURCE_POOL`, and `GOVC_FOLDER` need to be populated.
 
+Additionally, make sure `CONTROL_PLANE_ENDPOINT_IP` and `MGMT_CLUSTER_KUBECONFIG_PATH` are set.
```
ecpullen: Set with what?

bottlerocket/samples/scripts/create_sample_file.sh
```diff
@@ -0,0 +1,299 @@
+#!/bin/bash
+
+function create_ecs_migration_test() {
```
ecpullen: Is it possible to deduplicate any of the code? I had originally envisioned all of these env variables being set at the top of the Makefile so a user could overwrite them, and then the target would just do the variable substitution and run the yaml.

ecpullen (changes requested):

bottlerocket/samples/scripts/create_sample_file.sh
mjsterckx: That was my initial plan as well, but I spent the better part of yesterday trying to do just that and I was not able to figure out variable substitution inside a Makefile target. Both `eval` and `envsubst` do not work.

bottlerocket/samples/RUNBOOK.md
mjsterckx: I just left that as a comment for myself to remember to elaborate. I'll expand on that once I get around to updating the runbook, after the Makefile is done.

bottlerocket/samples/scripts/create_sample_file.sh
stmcginnis: In the makefile you should be able to do something like:

```bash
VARIANT=aws-k8s-1.24 ARCHITECTURE=x86_64 scripts/create_sample_file.sh sonobuoy-migration eks
```

mjsterckx: ^ That would still require having the script, which is fine by me, but I'm not sure if that aligns with ecpullen's vision. So now I'm thinking: export all variables at the top of the Makefile if not already set, and inside the target, just call the script with the two arguments, which does the variable substitution and writes the output file. The output file location is echo'd, so the Makefile target can run the test. @ecpullen Thoughts?

ecpullen: Ideally, the logic would be unified such that when a new sample is added, the new sample could be added to a `samples` list at the top of the Makefile (similar to how we add new bottlerocket agents) with nothing else (except additional env variables that aren't already in the Makefile). I am open to the implementation details if that is the end result.

mjsterckx: The problem is that as far as I can tell, each Makefile variable needs to be passed in explicitly to the create_script command. I changed the Makefile so it now contains all variables and the script doesn't have any duplicated code anymore, but it's still not exactly what you want. For example, it doesn't seem possible to set VARIANT if not already set (VARIANT ?= ...) inside of a make target, so the only option I can think of is to set VARIANT inside the bash script. Consequence of this is that if we add a new sample file, we'll have to add a case to the switch statement in the bash script as well as add a new target in the makefile that looks like the other targets currently in there.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+''']}
+
+[tasks.set-pseudo-env]
+dependencies = ["set-agent-images", "cluster-name", "gpu", "metadata-url", "ova-name", "bottlerocket-ami-id"]
+
+[tasks.set-env]
+dependencies = ["set-pseudo-env", "mgmt-cluster-kubeconfig-base64"]
+
+[tasks.set-env.env]
+CLUSTER_NAME = { value = "${PSEUDO_CLUSTER_NAME}", condition = { env_not_set = ["CLUSTER_NAME"] } }
```
ecpullen: All of these conditions should be when the `PSEUDO` are set, and the pseudos can be removed.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+        else
+            echo "[\"m6g.large\"]"
+        fi
+    fi
+else
+    echo ${INSTANCE_TYPES}
+fi
+''']}
+
+[tasks.create-inner]
```
ecpullen, line 189: Can you verify that `create-inner` is idempotent.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+script = "cli run file ${OUTPUT_FILE}"
+
+[tasks.run]
+script = '''
+#!@duckscript
+set_env TEST_TYPE ${1}
+cm_run_task ${1}
+cm_run_task create-inner
+cm_run_task run-inner
+'''
```
ecpullen:
```suggestion
[tasks.run]
dependencies = ["create"]
run_task="run-inner"
```

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+#!@duckscript
+set_env TEST_TYPE ${1}
+cm_run_task ${1}
+cm_run_task create-inner
+'''
+
+[tasks.install-controller]
+script = "cli install --controller-uri ${CONTROLLER_IMAGE_URI}"
+
+[tasks.add-secrets]
```
ecpullen: Let's break this into 2 targets. One for vSphere and one for AWS creds. You can use `conditions` to only have the vSphere secret be created for VMware variants.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+VARIANT = "aws-k8s-1.24-nvidia"
+
+[tasks.vmware-migration-test.env]
+VARIANT = "vmware-k8s-1.24"
+
+[tasks.vmware-sonobuoy-test.env]
+VARIANT = "vmware-k8s-1.24"
+
+[tasks.get-samples.env]
+SAMPLES = {script=["echo $(ls ${CLUSTER_TYPE})"]}
```
ecpullen, line 232:
```suggestion
SAMPLES = {script=["ls ${CLUSTER_TYPE}"]}
```
Does this work?

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+
+[tasks.unset-variables.env]
+CLUSTER_NAME = {unset = true}
```
ecpullen, line 235: If we are clearing this for all samples, should we allow a `CLUSTER_NAME_PREFIX`?

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,248 @@
...
+# GOVC_FOLDER = 
+
+[tasks.install-cli]
+script=['''
+cargo install --path "${CARGO_MAKE_WORKING_DIRECTORY}/../../cli"
+''']
+
+[tasks.set-agent-version]
+dependencies = ["install-cli"]
+env = { AGENT_IMAGE_VERSION = {script=["echo ${AGENT_IMAGE_VERSION:-$(cli --version | awk '{print $2}')}"]} }
```
ecpullen: This should be overridable by the user.

ecpullen (changes requested):
Great work!

bottlerocket/samples/Makefile.toml
mjsterckx: We're assigning the value in `PSEUDO_CLUSTER_NAME` to `CLUSTER_NAME`, if `CLUSTER_NAME` is not set in advance by the user. It works correctly when I test exporting `CLUSTER_NAME` before running the make command. Is my reasoning incorrect?

bottlerocket/samples/Makefile.toml
mjsterckx: That only gets the last filename for some reason. Using `ls -xm` to put commas in between the files only stores the last 2 in the variable. This is the most reliable way I've found that works.

bottlerocket/samples/Makefile.toml
ecpullen: I don't see why you can't just assign to `CLUSTER_NAME` instead of `PSEUDO_CLUSTER_NAME`.

mjsterckx: I couldn't use the `{ value = ..., condition = ...}` syntax because the value is being generated by some commands. Since I couldn't use that syntax, I had to unconditionally assign it to an environment variable, which would've overwritten the user-defined `CLUSTER_NAME`. That's why I assign it to a pseudo variable first, which is then conditionally passed into `CLUSTER_NAME` (only if the user didn't define that first).

bottlerocket/samples/Makefile.toml
mjsterckx: The `echo ${AGENT_IMAGE_VERSION:-...` makes it overridable, no?

bottlerocket/samples/Makefile.toml
mjsterckx: I'm not exactly happy with my current setup. The problem with running all files in a loop one after another is that all those variables (such as `CLUSTER_NAME`) are set after the first iteration and future iterations will take and keep that value. That's why I'm unsetting them after each iteration, but then if the user provided a value, that gets thrown away immediately. One solution to this would be to have the `run-samples` target create a separate `cargo make run <filename>` process per file, but I'd need other opinions to figure out if that's viable and desirable. I'm not sure about having `CLUSTER_NAME_PREFIX` as well as `CLUSTER_NAME`, it might cause confusion.

bottlerocket/samples/Makefile.toml
mjsterckx: ^ Ignore that, added conditional statements so that the PSEUDO variables could be removed.

bottlerocket/samples/Makefile.toml
ecpullen: I'm ok with it as is for now as long as it's documented somewhere.

bottlerocket/samples/Makefile.toml
ecpullen: Oh, I see. Can we do it the other way? It is way more readable imo.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+# CONTROL_PLANE_ENDPOINT_IP = 
+# MGMT_CLUSTER_KUBECONFIG_PATH = 
+# GOVC_USERNAME = 
+# GOVC_PASSWORD = 
+# GOVC_DATACENTER = 
+# GOVC_DATASTORE = 
+# GOVC_URL = 
+# GOVC_NETWORK = 
+# GOVC_RESOURCE_POOL = 
+# GOVC_FOLDER =
```
webern: ?

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+skip_core_tasks = true
+
+[env]
+AGENT_IMAGE_REGISTRY = { value = "public.ecr.aws/bottlerocket-test-system", condition = { env_not_set = ["AGENT_IMAGE_REGISTRY"] } }
+AGENTS = ["eks-resource-agent"]
+METADATA_BASE_URL = { value = "https://updates.bottlerocket.aws/2020-07-07", condition = { env_not_set = ["METADATA_BASE_URL"] } }
+ARCHITECTURE = { value = "x86_64", condition = { env_not_set = ["ARCHITECTURE"] } }
+CLUSTER_TYPE = { value = "eks", condition = { env_not_set = ["CLUSTER_TYPE"] } }
+ASSUME_ROLE = { value = "~", condition = { env_not_set = ["ASSUME_ROLE"] } }
+AWS_REGION = { value = "us-west-2", condition = { env_not_set = ["AWS_REGION"] } }
```
webern, line 11: Is it ok for these to be unprefixed? i.e. in Bottlerocket we use `BUILDSYS_` as a prefix for env vars. Will these clash with anything?

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+
+[tasks.cluster-name.env]
+CLUSTER_NAME = { script=["echo $(echo ${ARCHITECTURE} | tr '_' '-')-$(echo ${VARIANT} | tr -d '.')"], condition = { env_not_set = ["CLUSTER_NAME"] } }
+
+[tasks.ova-name.env]
+OVA_NAME = { script=["echo bottlerocket-${VARIANT}-x86_64-${STARTING_VERSION}.ova"], condition = { env_not_set = ["OVA_NAME"] } }
+
+[tasks.bottlerocket-ami-id.env]
+BOTTLEROCKET_AMI_ID = { script=['''
+if [ $(echo ${TEST_TYPE} | tr "-" " " | awk '{print $2}') = "migration" ]; then
```
webern: This looks brittle. Could we just use `STARTING_VERSION` if it's set, otherwise use latest?

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+fi
+aws ssm get-parameter \
+    --region ${AWS_REGION} \
+    --name "/aws/service/bottlerocket/${VARIANT}/$(echo ${ARCHITECTURE} | sed -e 's/aarch64/arm64/g')/${AMI_VERSION}/image_id" \
+    --query Parameter.Value --output text 
+'''], condition = { env_not_set = ["BOTTLEROCKET_AMI_ID"], env_contains = { "VARIANT" = "aws" } } }
+
+[tasks.gpu.env]
+GPU = { script=['''
+if [ $(echo ${VARIANT} | tr "-" " " | awk '{print $NF}') = "nvidia" ]; then
```
webern: :(

I get it though.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+'''
+
+[tasks.ecs-migration-test.env]
+VARIANT = "aws-ecs-1"
+
+[tasks.ecs-test.env]
+VARIANT = "aws-ecs-1"
+
+[tasks.ecs-workload-test.env]
+VARIANT = "aws-ecs-1-nvidia"
```
webern: Maybe you could set things like GPU and instance type here instead of the `awk`robatics.

ecpullen: I wanted the effort needed to add a sample to be the absolute minimum, which is why only the variant is set here. I think I prefer it this way.

webern: The effort to add a sample for a variant that ends with `-xyz` and is a workload test is higher this way.

bottlerocket/samples/Makefile.toml
mjsterckx: These are variables that need to be set by the user if a workload test file is being generated (the first two variables) or if a vmware file is being generated (the other variables) but I haven't found a way yet to convey that to the user. Might move them into a `help` target. This location is just temporary.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
+[config]
+skip_core_tasks = true
+
+[env]
+AGENT_IMAGE_REGISTRY = { value = "public.ecr.aws/bottlerocket-test-system", condition = { env_not_set = ["AGENT_IMAGE_REGISTRY"] } }
+AGENTS = ["eks-resource-agent"]
+METADATA_BASE_URL = { value = "https://updates.bottlerocket.aws/2020-07-07", condition = { env_not_set = ["METADATA_BASE_URL"] } }
+ARCHITECTURE = { value = "x86_64", condition = { env_not_set = ["ARCHITECTURE"] } }
```
ecpullen, line 8: nit: We use `ARCH` _everywhere_ else, so we probably should here.

bottlerocket/samples/Makefile.toml
ecpullen: @mjsterckx Can you add a comment at the top explaining that these are optional and provide comments how they should be filled.

bottlerocket/samples/Makefile.toml
mjsterckx: They won't be exported outside of the makefile process, so the only risk is that they were defined prior to running `cargo make` and contain values that will clash with the desired values for the generated sample. That was more of a discussion for the previous PRs though, where I added the sample files, so I'm hesitant to change it now. Either way, they won't have any side effects outside of the `cargo make` invocation.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,242 @@
...
+cm_run_task create-inner
+'''
+
+[tasks.install-controller]
+dependencies = ["set-agent-images"]
+script = "cli install --controller-uri ${CONTROLLER_IMAGE_URI}"
+
+[tasks.add-aws-secret]
+dependencies = ["install-controller"]
+condition = { env_contains = { "CLUSTER_TYPE" = "kind", "VARIANT" = "aws" } }
```
ecpullen: Will `"VARIANT" = "aws"` ever happen? We also need aws secrets for VMware migration testing. So I'd add the secret for all kind samples.

bottlerocket/samples/Makefile.toml
mjsterckx: I suppose this can be made to look cleaner by using a regex string compare (aka "contains nvidia") but the logic will be the same.

ecpullen: Let's use the regex just for clarity.

bottlerocket/samples/Makefile.toml
mjsterckx: @ecpullen wanted a separate target for each variable and I'm inclined to agree that it looks cleaner than putting several variables inside the `test` target. Leaving it the way it is now also makes it easier for others to add their own sample files and targets.

bottlerocket/samples/Makefile.toml
ecpullen: NVM. I see that it is looking for a substring here. Cool!

mjsterckx: It's an `env_contains` target so this checks if VARIANT contains the string `aws` (the syntax is a bit confusing). I'm not sure where the aws secret for vmware migration testing would be used, since all migration testing samples are in the `eks` folder. Do you mean just in case someone adds a migration test file that would be running on a `kind` cluster?

bottlerocket/samples/Makefile.toml
mjsterckx: That would mean I'd have to change it in all sample files as well. Not a problem, just want to confirm that that's what you want?

bottlerocket/samples/Makefile.toml
ecpullen: Just to be safe. Currently the migration agent requires aws secret.

bottlerocket/samples/Makefile.toml
ecpullen: I guess we can change it in a follow up pr.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,282 @@
...
+# EKS_RESOURCE_AGENT_IMAGE_URI = 
+# VSPHERE_K8S_CLUSTER_RESOURCE_AGENT_IMAGE_URI = 
+# VSPHERE_VM_RESOURCE_AGENT_IMAGE_URI = 
+# CONTROLLER_IMAGE_URI = 
+# METADATA_URL = 
+# CLUSTER_NAME = 
+# OVA_NAME = 
+# BOTTLEROCKET_AMI_ID = 
+# GPU = 
+# INSTANCE_TYPES =
```
ecpullen: I think we can get rid of this.

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,282 @@
...
+else
+    AMI_VERSION="latest"
+fi
+aws ssm get-parameter \
+    --region ${AWS_REGION} \
+    --name "/aws/service/bottlerocket/${VARIANT}/$(echo ${ARCHITECTURE} | sed -e 's/aarch64/arm64/g')/${AMI_VERSION}/image_id" \
+    --query Parameter.Value --output text 
+'''], condition = { env_not_set = ["BOTTLEROCKET_AMI_ID"], env_contains = { "VARIANT" = "aws" } } }
+
+[tasks.gpu.env]
```
ecpullen, line 89: We should probably check that `VARIANT` exists

bottlerocket/samples/Makefile.toml
```diff
@@ -0,0 +1,282 @@
...
+if [ -n "${STARTING_VERSION}" ]; then
+    AMI_VERSION=$(echo ${STARTING_VERSION} | tr -d "v")
+else
+    AMI_VERSION="latest"
+fi
+aws ssm get-parameter \
+    --region ${AWS_REGION} \
+    --name "/aws/service/bottlerocket/${VARIANT}/$(echo ${ARCHITECTURE} | sed -e 's/aarch64/arm64/g')/${AMI_VERSION}/image_id" \
+    --query Parameter.Value --output text 
+'''], condition = { env_not_set = ["BOTTLEROCKET_AMI_ID"], env_contains = { "VARIANT" = "aws" } } }
```
ecpullen: Can we put this condition for the task instead of the individual env variable?

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
- "ACCESS_KEY_ID=${ACCESS_KEY_ID}" \
- "SECRET_ACCESS_KEY=${SECRET_ACCESS_KEY}"
+A list of the customizable variables can be found at the top of [Makefile.toml](./Makefile.toml).
 
-eval "cat > ${OUTPUT_FILE} << EOF
-$(< kind/ecs-test.yaml)
-EOF
-" 2> /dev/null
-```
+_Note_: `ASSUME_ROLE` has a default value of `~` (null), but you can replace this with the ARN of an AWS IAM role that should be used for all AWS calls.
```
ecpullen: This note may need to be reworded `~`?

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
 
-### Workload Testing on `aws-ecs` Variants
+## Generating a test file
 
-_Note_: An example of a workload test that can be used with the `ecs-workload-agent` is the `nvidia-smoke` test. See the `nvidia-smoke` [README.md](../tests/workload/nvidia-smoke/README.md) for instructions on how to build and use its image.
+A test file can be generated by executing a `cargo make create <test-name>` command. The following example will create an ECS test file.
```
ecpullen: One sentence per line and `test-name` -> `sample-name`.
```suggestion
A test file can be generated by executing a `cargo make create <sample-name>` command. 
The following example will create an ECS test file.
```

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
-$(< kind/k8s-workload-test.yaml)
-EOF
-" 2> /dev/null
-```
-
-### Conformance Testing on `vmware-k8s` Variants
+### Workload tests
 
-This codeblock assumes that your vSphere config file has been sourced. Specifically, the variables `GOVC_USERNAME`, `GOVC_PASSWORD`, `GOVC_DATACENTER`, `GOVC_DATASTORE`, `GOVC_URL`, `GOVC_NETWORK`, `GOVC_RESOURCE_POOL`, and `GOVC_FOLDER` need to be populated.
+When creating or running a workload test file, the variables `WORKLOAD_TEST_NAME` and `WORKLOAD_TEST_IMAGE_URI` need to be set by the user.
```
ecpullen, line 55: Is there a check for this in the Makefile? Is there a way to check the sample for values that weren't replaced?

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
-  --region ${AWS_REGION} \
-  --name "/aws/service/bottlerocket/${VARIANT}/${ARCHITECTURE}/latest/image_id" \
-  --query Parameter.Value --output text)
+## Running a test file
 
-cli add-secret map  \
- --name "aws-creds" \
- "ACCESS_KEY_ID=${ACCESS_KEY_ID}" \
- "SECRET_ACCESS_KEY=${SECRET_ACCESS_KEY}"
+A test file can be run by executing a `cargo make run <test-name>` command. The following example will run an ECS test.
```
ecpullen: 1 line per sentence.

bottlerocket/samples/RUNBOOK.md
mjsterckx: A check to make sure that these variables have values can be added to the Makefile.

etungsten (approved):
Nice work figuring this all out. LGTM as long as outstanding comments are resolved.

> Created each sample with default values as well as overriding variables. Ran created samples on EKS and kind clusters.

Can you please show what the invocation and the resulting output looks like?

bottlerocket/samples/Makefile.toml
webern: I almost made the same comment. For better or worse we have standardized on `ARCH` in the Bottlerocket project and in other places within this project.

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
+Before running a `cargo make` command, several environment variables can be set to customize the generated file.
 
-_Note_: `ASSUME_ROLE` has a default value of `~` (null), but you can replace this with the ARN of an AWS IAM role that should be used for all AWS calls.
-
-The final `cat` command will print the populated file to the path indicated by `OUTPUT_FILE`.
-
-## EKS
+The directory from which to use the sample file corresponds with the value in `CLUSTER_TYPE`.
 
 The files in [eks](./eks) are meant to be run on an EKS test cluster. You can create a new cluster using the [eksctl](https://eksctl.io/introduction/) tool.
```
webern: I don't think we've used the term "management cluster" before, but maybe we should, to distinguish it from the target clusters.

```suggestion
The files in [eks](./eks) are meant to be run on an EKS management cluster. You can create a new cluster using the [eksctl](https://eksctl.io/introduction/) tool.
```

bottlerocket/samples/RUNBOOK.md
```diff
@@ -1,438 +1,57 @@
...
-### Conformance Testing on `vmware-k8s` Variants
+### Workload tests
 
-This codeblock assumes that your vSphere config file has been sourced. Specifically, the variables `GOVC_USERNAME`, `GOVC_PASSWORD`, `GOVC_DATACENTER`, `GOVC_DATASTORE`, `GOVC_URL`, `GOVC_NETWORK`, `GOVC_RESOURCE_POOL`, and `GOVC_FOLDER` need to be populated.
+When creating or running a workload test file, the variables `WORKLOAD_TEST_NAME` and `WORKLOAD_TEST_IMAGE_URI` need to be set by the user.
 
-```bash
-CONTROL_PLANE_ENDPOINT_IP=
-MGMT_CLUSTER_KUBECONFIG_PATH=
+_Note_: An example of a workload test that can be used with the `ecs-workload-agent` or `k8s-workload-agent` is the `nvidia-smoke` test. See the `nvidia-smoke` [README.md](../tests/workload/nvidia-smoke/README.md) for instructions on how to build and use its image.
```
webern: nit
```suggestion
_Note_: An example of a workload test that can be used with the `ecs-workload-agent` or `k8s-workload-agent` is the `nvidia-smoke` test.
See the `nvidia-smoke` [README.md](../tests/workload/nvidia-smoke/README.md) for instructions on how to build and use its image.
```

webern (approved):

bottlerocket/samples/Makefile.toml
mjsterckx: It'll be changed in a follow-up PR. I'd rather not overload this one by changing 12 more files.

bottlerocket/samples/Makefile.toml
mjsterckx: The `aws ssm` command will return exit code 1 and show the error if the query failed (which would indicate most likely that `VARIANT` does not exist). I'm not sure if any other checks are needed.

ecpullen (approved):
Awesome! 🎉

PR #754, opened 2023-01-31
Remove `bottlerocket/testsys` cli

ecpullen:
**Issue number:**

Closes #438 

**Description of changes:**

Removes the legacy cli from the project.

**Testing done:**

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #757, opened 2023-02-01
docs: updated QUICKSTART to use cli instead of testsys

mjsterckx:
**Issue number:**

N/A

**Description of changes:**

Updated `QUICKSTART.md` to use the `cli` CLI instead of the old `testsys` CLI since it is being removed in https://github.com/bottlerocket-os/bottlerocket-test-system/pull/754.

**Testing done:**

Ran through the new instructions, letting the test complete and pass.

**Terms of contribution:**

### DIFF
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -21,14 +21,14 @@ export EKS_REGION="us-west-2"
 export CARGO_HOME="${TESTSYS_DIR}/.cargo"
 K8S_VER="1.21"
 REGION="${EKS_RESION}"
-alias testsys="${TESTSYS_DIR}/.cargo/bin/testsys"
+alias cli="${TESTSYS_DIR}/.cargo/bin/cli"
 ```
 
-Install the `testsys` command line tool into the local CARGO_HOME and build the containers we need:
+Install the `cli` command line tool into the local CARGO_HOME and build the containers we need:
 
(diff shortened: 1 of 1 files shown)


etungsten (approved):

docs/QUICKSTART.md
```diff
@@ -60,81 +60,75 @@ Next we install the TestSys namespace, controller and CRD schemas into the TestS
...
+```
+
+This will delete all tests and resources where `destructionPolicy` is `onDeletion`.
+
+If you used `destructionPolicy: never`, there will be an EKS cluster running at the end of execution.
 This can be very convenient to keep around for subsequent test runs.
 
 It does however consume resources and incur usage charges.
-If you are done running tests and would like to clean up, run the following to delete the EKS cluster.
+If you are done running tests and would like to clean up the EKS cluster, run the following to delete it.
```
ecpullen, line 131: Don't the samples automatically clean up resources?

mjsterckx: Yeah, this follows line 127 where `destructionPolicy: never` is specified in case the user manually changed it themselves. I can remove that line but I thought it would be a good idea to leave it in in case the user wanted to keep a cluster around and couldn't figure out how to delete it afterwards.

ecpullen (approved):
LGTM

PR #770, opened 2023-02-09
TestManager: Ensure no resources are in the cluster before deleting the namespace

ecpullen:
**Issue number:**

N/A

**Description of changes:**

```
Author: ecpullen <pullenep@amazon.com>
Date:   Thu Feb 9 18:06:32 2023 +0000

    cli: Print underlying errors with anyhow
```
Show the underlying error when a command fails.

```
Author: ecpullen <pullenep@amazon.com>
Date:   Thu Feb 9 17:57:00 2023 +0000

    TestManager: Ensure no resources for uninstall
    
    If a resource is present in the cluster while the namespace is being
    deleted, there is a chance the resource deletion fails and puts the
    cluster in a bad state. To prevent that, all resources must be deleted
    before the namespace is deleted.
```

**Testing done:**

`cli install` followed by `cli uninstall`

With resources:
```
$ cli uninstall
Unable to uninstall testsys from the cluster. (Some artifacts may be left behind)

Caused by:
    Some resources are still in the cluster
```

After the resources are cleaned up:
```
$ cli uninstall
testsys components were successfully uninstalled.
```

**Terms of contribution:**

### DIFF
--- a/model/src/test_manager/error.rs
+++ b/model/src/test_manager/error.rs
@@ -50,6 +50,9 @@ pub enum Error {
     #[snafu(display("Unable to find {}", what))]
     NotFound { what: String },
 
+    #[snafu(display("Some resources are still in the cluster"))]
+    ResourceExisting,
+
     #[snafu(display("Unable to send event: {}", source))]
     Sender {
         source: futures::channel::mpsc::SendError,
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -68,7 +68,7 @@ async fn main() {
     let args = Args::parse();
     init_logger(args.log_level);
     if let Err(e) = run(args).await {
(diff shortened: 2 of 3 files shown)


model/src/test_manager/error.rs
```diff
@@ -50,6 +50,9 @@ pub enum Error {
     #[snafu(display("Unable to find {}", what))]
     NotFound { what: String },
 
+    #[snafu(display("Some resources were still in the cluster"))]
```
stmcginnis: nit: I would say "Some resources are still in the cluster".

stmcginnis (approved):
Looks good to me. One minor thing pointed out, but I think it's fine as is.

etungsten (approved):

PR #775, opened 2023-02-27
docs: added documentation for minimal iam permissions

### LINKED ISSUE #417: minimal set of iam permissions

mjsterckx:
**Issue number:**

Related: #417 

**Description of changes:**

Added a `bottlerocket/docs/permissions` folder that stores one or more JSON files for each agent. The files contain the minimal IAM permissions needed to create and run the test/resource. Added a `MINIMAL-PERMISSIONS.md` file that maps each sample test file in `bottlerocket/samples` to the corresponding minimal IAM permissions.

The JSON files were generated by running the sample test files with a different assume_role per test/resource. These assume_roles had `Administrator` permissions, and using CloudTrail, it was possible to see which permissions were actually used by that role from the time of creation to the time of deletion.

**Testing done:**

N/A

**Terms of contribution:**


ecpullen:
Looks good so far.

ecpullen (approved):

stmcginnis (approved):
This looks good to me.

PR #784, opened 2023-03-06
Samples: Use latest AMIs for testing

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Samples should use the latest Bottlerocket AMIs for testing instead of a set version. 

**Testing done:**

`cargo make create ecs-migration-test`

Verified that a 1.12.0 ami was selected and the migrations go:
* 1.12.0 -> 1.11.1
* 1.11.1 -> 1.12.0

**Terms of contribution:**

### DIFF
--- a/bottlerocket/samples/Makefile.toml
+++ b/bottlerocket/samples/Makefile.toml
@@ -9,7 +9,14 @@ CLUSTER_TYPE = { value = "eks", condition = { env_not_set = ["CLUSTER_TYPE"] } }
 ASSUME_ROLE = { value = "~", condition = { env_not_set = ["ASSUME_ROLE"] } }
 AWS_REGION = { value = "us-west-2", condition = { env_not_set = ["AWS_REGION"] } }
 UPGRADE_VERSION = { value = "v1.11.1", condition = { env_not_set = ["UPGRADE_VERSION"] } }
-STARTING_VERSION = { value = "v1.11.0", condition = { env_not_set = ["STARTING_VERSION"] } }
+STARTING_VERSION = { script = ["""\
+        aws ssm get-parameter \
+            --region us-west-2 \
+            --name "/aws/service/bottlerocket/aws-k8s-1.24/arm64/latest/image_version" \
+            --query Parameter.Value \
+            --output text \
+        | awk -F- '{printf "v%s", $1}' \
+    """], condition = { env_not_set = ["STARTING_VERSION"] } }
 SONOBUOY_MODE = { value = "quick", condition = { env_not_set = ["SONOBUOY_MODE"] } }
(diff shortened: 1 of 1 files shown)


mjsterckx (approved):
Thank you!

stmcginnis (approved):
Looks great!

bottlerocket/samples/Makefile.toml
```diff
@@ -9,7 +9,13 @@ CLUSTER_TYPE = { value = "eks", condition = { env_not_set = ["CLUSTER_TYPE"] } }
 ASSUME_ROLE = { value = "~", condition = { env_not_set = ["ASSUME_ROLE"] } }
 AWS_REGION = { value = "us-west-2", condition = { env_not_set = ["AWS_REGION"] } }
 UPGRADE_VERSION = { value = "v1.11.1", condition = { env_not_set = ["UPGRADE_VERSION"] } }
-STARTING_VERSION = { value = "v1.11.0", condition = { env_not_set = ["STARTING_VERSION"] } }
+STARTING_VERSION = { script = ["""\
+        aws ssm get-parameter \
+            --region us-west-2 \
+            --name "/aws/service/bottlerocket/aws-k8s-1.24/arm64/latest/image_version" \
```
gthao313, line 15: Is the `aws-k8s-1.24` what we target to use on testsys? I was thinking if we deprecate `aws-k8s-1.24` later, will we need update here?

gthao313 (approved):

bottlerocket/samples/Makefile.toml
ecpullen: Yes we use `aws-k8s-1.24` for all of our eks testing.

webern (approved):

webern:
It's a little sneaky as it is the only command that now requires the aws CLI to be installed, but that's fine.

bottlerocket/samples/Makefile.toml
```diff
@@ -9,7 +9,13 @@ CLUSTER_TYPE = { value = "eks", condition = { env_not_set = ["CLUSTER_TYPE"] } }
...
+            --query Parameter.Value \
+        | sed -rn 's/^"([0-9]+\\.[0-9]+\\.[0-9]+)-.*/v\\1/p' \
```
jpculp:
```suggestion
        aws ssm get-parameter \
            --region us-west-2 \
            --name "/aws/service/bottlerocket/aws-k8s-1.24/arm64/latest/image_version" \
            --query Parameter.Value \
            --output text
        | awk -F- '{printf "v%s", $1}' \
```

What you have isn't wrong, but this might be easier to read.

mjsterckx:
> It's a little sneaky as it is the only command that now requires the aws CLI to be installed, but that's fine.

We already needed to have the aws CLI installed before because we used it in the `bottlerocket-ami-id` target.

PR #787, opened 2023-03-07
EC2: Set security group to `None` if an empty `Vec` is provided

ecpullen:
**Issue number:**

N/A

**Description of changes:**

A n ew rust sdk version introduced a bug where we could no longer add an empty list of security groups and instead needed to add `None`.

**Testing done:**

Ran `cargo make test` with the updated EC2 agent and verified that the instances were launched.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
+++ b/bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
@@ -245,7 +245,11 @@ where
                 .min_count(instance_count)
                 .max_count(instance_count)
                 .subnet_id(subnet_id)
-                .set_security_group_ids(Some(spec.configuration.security_groups.clone()))
+                .set_security_group_ids(if spec.configuration.security_groups.is_empty() {
+                    None
+                } else {
+                    Some(spec.configuration.security_groups.clone())
+                })
                 .image_id(&spec.configuration.node_ami)
                 .instance_type(InstanceType::from(instance_type.as_str()))
                 .tag_specifications(tag_specifications(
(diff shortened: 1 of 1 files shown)


mjsterckx (approved):

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -245,7 +245,11 @@ where
                 .min_count(instance_count)
                 .max_count(instance_count)
                 .subnet_id(subnet_id)
-                .set_security_group_ids(Some(spec.configuration.security_groups.clone()))
+                .set_security_group_ids(if spec.configuration.security_groups.is_empty() {
+                    None
```
jpmcb, line 249: Is this a good idea? I'd like to understand more about this. From reading the code, this sounds like when we initiate the `ec2_client`, we are ok attaching no security groups to it. My instinct is that we should instead fail if there are no security groups since we'd likely always want to have a sane default.

ecpullen: This is a part of the `run_instances` function. When we run instances on an EKS cluster, we expect there to be security groups, but to run instances on an ECS cluster we do not expect security groups. Since security groups aren't required, we use `None` as the default.

webern (approved):

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
jpmcb: Makes sense!!! 👏

Issue #390, opened 2022-05-04
update aws-iam-authenticator to 0.5.5 or higher


webern:
There is a fairly substantial breaking change in 0.5.4. I tried and it breaks our eks-provider. We might have to replace v1alpha1 with v1beta1 in our kubeconfigs or something.

- https://stackoverflow.com/a/71319893/2779792
- https://github.com/kubernetes-sigs/aws-iam-authenticator/pull/386

webern:
Basically when updating to a newer aws-iam-authenticator, I found it to be incompatible with the kubeconfig files we were getting from eksctl.

PR #805, opened 2023-04-06
Makefile: aws-iam-authenticator 0.6.8

jpmcb:
**Issue number:**

N/a - related to GitHub rate limiting on `go mod vendor`-ing all the code within this image.

**Description of changes:**

```
Makefile: aws-iam-authenticator 0.6.8

- Skip vendoring kubernetes code explicitly
```

This patch is twofold:
- Adds the new aws-iam-authenticator in the bottlerocket-lookaside-cache
  (which mitigates issues building the older one with `direct` as the goproxy
  setting - see: https://github.com/kubernetes-sigs/aws-iam-authenticator/issues/586)
- Skips vendoring code from `kubernetes/kubernetes`: the upstream project 
  _already_ vends it's dependencies, so all we need to do is pass the `vendor`
  directory to the license checker. My theory on the rate limiting is that since
  the docker container is using `https` to pull code through `git`, we are 
  getting rate limited quickly. But since the majority of that code is in
  kubernetes, we can skip this and avoid getting rate limited

**Testing done:**

Able to `make tools` without flakes.

**Terms of contribution:**


ecpullen (changes requested):
Blocking until a full test has been done to ensure none of our agents are broken.
#390

jpmcb:
The other alternative is to add a flip on `GOPROXY` to also use the upstream proxy and not just `direct`

jpmcb:
Superseded by #810

PR #844, opened 2023-06-30
Remove workaround for eksctl build

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

There was a release of eksctl that would fail due to a retagged dependency causing a hash mismatch. To get past this we added some handling to update the go.mod/go.sum files when building eksctl as part of our tools build.

eksctl has now moved past that broken version of the dependency, and the current version we use no longer needs this workaround, so this cleans up that extra handling.

**Testing done:**

`make tools`

**Terms of contribution:**

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -44,12 +44,6 @@ RUN mkdir eksctl && \
     rm "eksctl_${EKSCTL_VERSION}.tar.gz"
 
 WORKDIR /home/builder/eksctl/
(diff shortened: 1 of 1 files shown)


ecpullen (approved):

webern (approved):

stmcginnis:
Someone with more repo rights may need to merge this. For some reason it thinks it needs the results from `images`, but that workflow ignores this `tools/**` path, so that's never going to happen.

PR #858, opened 2023-09-07
Build eksctl-anywhere binary from source

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

Existing method gets eksctl-anywhere by curling the binary of a specific release version. Base functionality is fairly stable, and we often need to pick up the latest when adding a new variant to get support for a newer k8s version. That is the case now with needing the k8s 1.28 support that is not in a release yet.

**Testing done:**

Ran `make images`, built with no error.

```sh
$ TESTSYS_RESOURCE_ACTION=test docker run -it --rm metal-k8s-cluster-resource-agent:latest bash
bash-4.2# eksctl version
0.144.0
bash-4.2# eksctl-anywhere version
v0.0.0-dev-release-0.17
```

Cleaned up caching and run `make images -e TESTSYS_EKSA_REF=heads/main` to pick up the latest `eksctl-anywhere` from `main`.

```sh
$ TESTSYS_RESOURCE_ACTION=test docker run -it --rm metal-k8s-cluster-resource-agent:latest bash
bash-4.2# eksctl-anywhere version
v0.0.0-dev
```

**Terms of contribution:**


Dockerfile
```diff
@@ -154,6 +154,8 @@ ENTRYPOINT ["./ecs-resource-agent"]
 # Builds the vSphere K8s cluster resource agent image
 FROM public.ecr.aws/amazonlinux/amazonlinux:2 as vsphere-k8s-cluster-resource-agent
 
+ENV K8S_1_28_SUPPORT=true
+
```
etungsten: I think it would be best to let the testsys agent themselves determine the env var to populate its environment with.

tools/Dockerfile
```diff
@@ -93,12 +85,7 @@ RUN cp -p LICENSE /usr/share/licenses/eksctl-anywhere && \
       --spdx-data /usr/libexec/tools/spdx-data \
       --out-dir /usr/share/licenses/eksctl-anywhere/vendor \
       go-vendor ./vendor
-RUN curl -L "${EKSA_BINARY_URL}" \
-      -o "eksctl_anywhere_${EKSA_VERSION}_${EKSA_RELEASE_NUMBER}_${GOOS}_amd64.tar.gz" && \
-    grep "eksctl_anywhere_${EKSA_VERSION}_${EKSA_RELEASE_NUMBER}_${GOOS}_amd64.tar.gz" \
-      /hashes/eksctl-anywhere | sha512sum --check - && \
```
etungsten, line 99: We would no longer need the associated hash in https://github.com/bottlerocket-os/bottlerocket-test-system/blob/develop/tools/hashes/eksctl-anywhere

tools/Dockerfile
```diff
@@ -65,24 +65,16 @@ USER root
 RUN mkdir -p /usr/share/licenses/eksctl-anywhere && \
     chown -R builder:builder /usr/share/licenses/eksctl-anywhere
 
-ARG EKSA_VERSION=0.16.0
-ARG EKSA_SOURCE_URL="https://github.com/aws/eks-anywhere/archive/refs/tags/v${EKSA_VERSION}.tar.gz"
+# Build the latest EKS-A from main (we may want to change this to get a specific hash)
+ARG EKSA_SOURCE_URL="https://github.com/aws/eks-anywhere/archive/refs/heads/main.tar.gz"
```
etungsten: I think we should keep some way to change the revision we want to build from. Default to some stable released version.

stmcginnis: > Default to some stable released version.

I may have misunderstood, but I thought we could not set this to the released version. At least for now while we are trying to pick up k8s 1.28 support.

I can update this to use a variable, but please clarify what you think it should be pulling.

etungsten: For regular releases of testsys images, we should still be building EKS-A from stable release versions. We want to allow the possibility of overriding the variable so we can build from any arbitrary git revision for special testing, e.g. new K8s version testing.

etungsten: This would also ideally mean we need some way of specifying that through `make` and plumbing it to `docker build`

etungsten (commented):
LGTM if testing is good.

To test I think we can first make sure it builds, then try `docker run`ning a shell in `vsphere-k8s-cluster-resource-agent` and check the version of `eksctl-anywhere`.

stmcginnis:
Build now passing, with the exception of the existing failure for the `Build/build` check.

webern:
What's going on with the CI failures for --offline?

stmcginnis:
> What's going on with the CI failures for --offline?

I don't know, I've asked that on several PRs over the last month or two, but no one has given a good answer. I tried digging in a little, but I was not able to figure out what changed to cause this to start happening.

My guess it is either due to a change in the SDK or a change in the newer versions of the Rust toolchain used. It seems like something isn't getting mounted in the right place for the `--offline` build to find the cached crates.

tools/Dockerfile
arnaldo2792: We could still do hash validations. If the source was modified, an additional parameter can be given with the new expected hash.

arnaldo2792 (approved):

etungsten (approved):

PR #882, opened 2024-01-12
controller: Add cloudwatch logging for pods

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Add an optional flag to the controller for archiving pod logs with CloudWatch.

**Testing done:**

Verified that logs are archived when the flag is set and not if the flag is not set. Tested archival for completed pods and interrupted pods.

**Terms of contribution:**

### DIFF
--- a/controller/src/job/error.rs
+++ b/controller/src/job/error.rs
@@ -13,14 +13,42 @@ pub(crate) enum JobError {
     #[snafu(display("Unable to create job: {}", source))]
     Create { source: kube::Error },
 
+    #[snafu(display("Unable to create log event '{}': {:?}", log_event, source))]
+    CreateLogEvent {
+        log_event: String,
+        source: aws_sdk_cloudwatchlogs::types::SdkError<
+            aws_sdk_cloudwatchlogs::error::PutLogEventsError,
+        >,
+    },
+
+    #[snafu(display("Unable to create log group '{}': {}", log_group, message))]
+    CreateLogGroup { log_group: String, message: String },
+
+    #[snafu(display("Unable to create log stream '{}': {:?}", log_stream, source))]
+    CreateLogStream {
+        log_stream: String,
+        source: aws_sdk_cloudwatchlogs::types::SdkError<
+            aws_sdk_cloudwatchlogs::error::CreateLogStreamError,
+        >,
+    },
+
     #[snafu(display("Unable to delete job: {}", source))]
     Delete { source: kube::Error },
 
     #[snafu(display("Unable to get job: {}", source))]
     Get { source: kube::Error },
 
+    #[snafu(display("Unable to read logs for pod '{}': {}", pod, source))]
+    NoLogs { pod: String, source: kube::Error },
+
+    #[snafu(display("No pods found for job '{}'", job))]
+    NoPods { job: String },
+
     #[snafu(display("Job does not exist: {}", source))]
     NotFound { source: kube::Error },
+
+    #[snafu(display("{}", source), context(false))]
+    SystemTime { source: std::time::SystemTimeError },
 
     #[snafu(display(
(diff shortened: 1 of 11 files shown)


controller/src/job/error.rs
```diff
@@ -13,15 +13,43 @@ pub(crate) enum JobError {
     #[snafu(display("Unable to create job: {}", source))]
     Create { source: kube::Error },
 
+    #[snafu(display("Unable to create log event '{}': {:?}", log_event, source))]
+    CreateLogEvent {
+        log_event: String,
+        source: aws_sdk_cloudwatchlogs::types::SdkError<
```
rpkelly, line 19: Just to confirm, since we've run into it before, will this unwrap the error fully to something useful?

controller/src/job/mod.rs
```diff
@@ -122,3 +127,116 @@ pub(crate) async fn delete_job(k8s_client: kube::Client, name: &str) -> JobResul
...
+        .first()
+        .context(error::NoPodsSnafu {
+            job: job_name.to_string(),
+        })?
+        .name_any();
+
+    Ok(name)
+}
+
+async fn pod_logs(k8s_client: kube::Client, pod_name: &str) -> JobResult<String> {
```
rpkelly, line 150: Doesn't the testsys model already have a way to get logs? Why do we need a new, different way here?

controller/src/job/mod.rs
```diff
@@ -122,3 +127,116 @@ pub(crate) async fn delete_job(k8s_client: kube::Client, name: &str) -> JobResul
...
+        .create_log_group()
+        .log_group_name("testsys")
+        .send()
+        .await
+    {
+        Ok(_) => info!("Creating log group"),
+        Err(e) => {
+            let service_error = e.into_service_error();
+            if service_error.is_resource_already_exists_exception() {
+                info!("Log group already exists.")
```
webern, line 192: Maybe a little noisy?

```suggestion
                debug!("testsys Cloudwatch Log group already exists.")
```

controller/src/job/mod.rs
```diff
@@ -122,3 +127,116 @@ pub(crate) async fn delete_job(k8s_client: kube::Client, name: &str) -> JobResul
...
+    };
+    let pod_api: Api<Pod> = Api::namespaced(k8s_client, NAMESPACE);
+
+    pod_api
+        .logs(pod_name, &log_params)
+        .await
+        .context(error::NoLogsSnafu { pod: pod_name })
+}
+
+pub(crate) async fn archive_logs(k8s_client: kube::Client, job_name: &str) -> JobResult<()> {
```
webern, line 164: when does this happen? Is it only after the test has completed or failed? Or can this be kicked off at the start of the pod and the logs stream in realtime?

ecpullen: It happens as a job is deleted. I wasn't able to find a solution for realtime logs.

webern: It might not be something "inside" the system, but there are likely plenty of solutions for sending pod logs in realtime, so we should consider using those as well. Not going to block the PR though as we need to get something going for this.

webern (approved):

controller/src/job/error.rs
ecpullen: Yes it will show the underlying error since it is using debug.

controller/src/job/mod.rs
ecpullen: We are accessing the logs from different interfaces. (ResourceInterface vs TestManager)

rpkelly: We should backlog an issue to unify this so that log fetching updates do not require updating 2 places. Not necessary for this change though

rpkelly (approved):

PR #883, opened 2024-01-16
sonobuoy: Delete namespace before attempting test

ecpullen:
**Issue number:**

**Description of changes:**

Automatically delete the sonobuoy namespace to prevent it from causing test errors.

**Testing done:**

Built the sonobuoy agent and verified that it worked with both an existing namespace and without a namespace.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
+++ b/bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
@@ -86,6 +86,14 @@ where
             .await?;
         info!("Stored kubeconfig in {}", TEST_CLUSTER_KUBECONFIG_PATH);
 
+        match delete_workload(TEST_CLUSTER_KUBECONFIG_PATH).await {
+            Ok(_) => {}
+            Err(e) => info!(
+                "Unable to delete workload testing namespace. It is possible it was already deleted. {}",
+                e
+            ),
+        };
+
         run_workload(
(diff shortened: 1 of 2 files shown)


bottlerocket/agents/src/bin/sonobuoy-test-agent/main.rs
```diff
@@ -94,6 +94,14 @@ where
             None => None,
         };
 
+        match delete_sonobuoy(TEST_CLUSTER_KUBECONFIG_PATH).await {
+            Ok(_) => {}
+            Err(e) => warn!(
```
webern: I think this is good, but I think this should be `info!` and not `warn!` since it probably happens this way more often than not.

webern (approved):

bottlerocket/agents/src/bin/k8s-workload-agent/main.rs
```diff
@@ -86,6 +86,14 @@ where
             .await?;
         info!("Stored kubeconfig in {}", TEST_CLUSTER_KUBECONFIG_PATH);
 
+        match delete_workload(TEST_CLUSTER_KUBECONFIG_PATH).await {
+            Ok(_) => {}
+            Err(e) => warn!(
+                "Unable to delete sonobuoy namespace. It is possible it was already deleted. {}",
```
rpkelly: Should this say sonobouy or workload?

gthao313 (approved):

PR #889, opened 2024-02-02
controller: Prevent existing log group from stopping log retention

ecpullen:
**Issue number:**
N/A

**Description of changes:**

If the log group `testsys` already exists in an account, then the controller errors because it is not able to create the log group. Instead, the controller should still try to archive the logs.

**Testing done:**

- [ ] Test the controller without testsys log group existing
- [ ] Test the controller with the testsys log group existing

**Terms of contribution:**

### DIFF
--- a/controller/src/job/mod.rs
+++ b/controller/src/job/mod.rs
@@ -9,7 +9,7 @@ use k8s_openapi::api::core::v1::Pod;
 use k8s_openapi::chrono::{Duration, Utc};
 use kube::api::{DeleteParams, ListParams, LogParams, PropagationPolicy};
 use kube::{Api, ResourceExt};
-use log::{debug, info, warn};
+use log::{debug, info, trace, warn};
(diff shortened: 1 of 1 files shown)


controller/src/job/mod.rs
```diff
@@ -206,15 +206,16 @@ pub(crate) async fn archive_logs(k8s_client: kube::Client, job_name: &str) -> Jo
...
         .log_group_name("testsys")
         .log_stream_name(&name)
         .send()
         .await
-        .context(error::CreateLogStreamSnafu {
-            log_stream: name.to_string(),
-        })?;
+        .is_err()
+    {
+        trace!("Unable to create log group it is possible it already exists.");
```
webern, line 217: A couple of things, you should log the error with `{:?}`.
You should check for the specific "it already exists" error condition and only ignore that, otherwise issues like authorization, etc could be difficult to diagnose.

PR #919, opened 2024-07-18
Update dependencies, tighten cargo-deny settings

cbgbt:
**Description of changes:**
> Run cargo-deny using the SDK
---
> clippy: fixes for rust version update to 1.78
---
> build using the Bottlerocket SDK
---
> Update sdk to v0.42.0
---
> Update kube version to 0.88 & k8s-openapi to 0.21
---
> Update AWS SDK for Rust to 1
---
> cargo update, tighten cargo-deny settings
>
> This change also contains several major-version updates for dependencies to pass the cargo-deny checks.

**Testing done:**
* [x] integration testing

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,8 +22,7 @@ jobs:
       labels: bottlerocket_ubuntu-latest_8-core
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.72 && rustup component add rustfmt && rustup component add clippy
-      - run: make build
+      - run: make sdk-build
         # If we forget to add yamlgen changes to our commits, this will fail.
       - name: ensure that git is clean
         run: test -z "$(git status --untracked-files=all --porcelain)"
@@ -50,5 +49,4 @@ jobs:
     if: ${{ github.event_name == 'pull_request' }}
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.72
       - run: make cargo-deny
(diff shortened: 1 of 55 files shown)


jpculp:
This is great! For the crate dependencies, I found we can reduce some duplication with the following changes:
- `base64`: `pem` v3.0.4 → v3.0.3, and setting our cargo.tomls to `base64 0.21`
- `ring`: `sct` v0.7.0 → v0.7.1

cbgbt:
The integration tests run kind, which creates a cluster within Docker. Since I changed these tests to be run within the SDK container, they failed without access to the docker socket.

This change modifies the integration test runner to have access to the docker socket, plus network access to interact with the kind cluster running in an additional container.

cbgbt:
> This is great! For the crate dependencies, I found we can reduce some duplication with the following changes:

@jpculp  Nice wrangling, though since these dependencies are smaller, I think may prefer to keep them more up-to-date over eliminating dupes. What do you think?

jpculp:
> @jpculp Nice wrangling, though since these dependencies are smaller, I think may prefer to keep them more up-to-date over eliminating dupes. What do you think?

I'm not super worried about `base64`, but with `pem` the only difference between 3.0.3 and 3.0.4 is the `base64`. If we want the benefits of `base64` 0.22 over 0.21 then yeah go for it.

As far as `sct` goes, there is no reason not to update to `0.7.1`.

cbgbt:
Good point! Force pushed for the dependency changes mentioned by @jpculp.

jpculp (approved):

sumukhballal (approved):

cbgbt:
I'm convinced that we can update our `aws-config` `BehaviorVersion` based on the following:
* https://docs.rs/aws-config/1.5.4/aws_config/struct.BehaviorVersion.html
* https://github.com/awslabs/tough/pull/781/commits/3a55041e00311dccc0f0b4ab6e72615fbc9126af

PR #954, opened 2024-12-02
ipv6 cluster creation issue in testsys

zongzhidanielhu:
**Issue number:**
ipv6 variants always have a ipFamily=ipv4
   

**Description of changes:**
changed the way of create cluster so that ipv6 has ipFamily=ipv6, ipv4 has ipFamily=ipv4

**Testing done:**
1. after codebase changed, but new docker image for bottlerocket-test-system
2. push image into ECR
3. trigger the changes in BoB, the applied variant is aws-k8s-1.24
4. Checked cluster status in cloudFormation, the aforementioned issue has been solved

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -20,16 +20,19 @@ use resource_agent::provider::{
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::env::temp_dir;
-use std::fs::write;
 use std::path::Path;
 use std::process::Command;
 use testsys_model::{Configuration, SecretName};
+
+use std::fs::File;
+use std::io::Write;
 
 /// The default region for the cluster.
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
 const DEFAULT_VERSION: &str = "1.24";
 const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
+const CLUSTER_CONFIG_PATH: &str = "/local/cluster_config.yaml";
 
 #[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -115,7 +118,6 @@ impl AwsClients {
         }
     }
 }
-
 enum ClusterConfig {
     Args {
         cluster_name: String,
@@ -127,6 +129,123 @@ enum ClusterConfig {
         cluster_name: String,
         region: String,
     },
+}
+enum IPFamily {
+    IPv4,
+    IPv6,
+}
+
+fn get_family_string(family: IPFamily) -> &'static str {
+    match family {
+        IPFamily::IPv4 => "IPv4",
+        IPFamily::IPv6 => "IPv6",
+    }
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Config {
+    api_version: String,
+    kind: String,
+    metadata: Metadata,
+    availability_zones: Vec<String>,
+    kubernetes_network_config: KubernetesNetworkConfig,
+    addons: Vec<Addon>,
+    iam: IAMConfig,
+    managed_node_groups: Vec<ManagedNodeGroup>,
+}
+
+#[derive(Serialize)]
+struct Metadata {
+    name: String,
+    region: String,
+    version: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct KubernetesNetworkConfig {
+    ip_family: String,
+}
+
+#[derive(Serialize)]
+struct Addon {
+    name: String,
+    version: String,
+}
+
+#[derive(Serialize)]
+#[allow(non_snake_case)]
+struct IAMConfig {
+    withOIDC: bool,
+}
+
+#[derive(Serialize)]
+struct ManagedNodeGroup {
+    name: String,
+}
+
+#[allow(clippy::unwrap_or_default)]
+fn create_yaml(
+    cluster_name: &str,
+    region: &str,
+    version: &str,
+    zones: &Option<Vec<String>>,
+) -> ProviderResult<()> {
+    let ip_family = if cluster_name.ends_with("ipv6") {
+        get_family_string(IPFamily::IPv6)
+    } else {
+        get_family_string(IPFamily::IPv4)
+    };
+
+    let cluster = Config {
+        api_version: "eksctl.io/v1alpha5".to_string(),
+        kind: "ClusterConfig".to_string(),
+        metadata: Metadata {
+            name: cluster_name.to_string(),
+            region: region.to_string(),
+            version: version.to_string(),
+        },
+        availability_zones: zones.clone().unwrap_or_else(Vec::new),
+        kubernetes_network_config: KubernetesNetworkConfig {
+            ip_family: ip_family.to_string(),
+        },
+        addons: vec![
+            Addon {
+                name: "vpc-cni".to_string(),
+                version: "latest".to_string(),
+            },
+            Addon {
+                name: "coredns".to_string(),
+                version: "latest".to_string(),
+            },
+            Addon {
+                name: "kube-proxy".to_string(),
+                version: "latest".to_string(),
+            },
+        ],
+        iam: IAMConfig { withOIDC: true },
+        managed_node_groups: vec![ManagedNodeGroup {
+            name: "mng-1".to_string(),
+        }],
+    };
+
+    let yaml =
+        serde_yaml::to_string(&cluster).context(Resources::Clear, "Failed to serialize YAML");
+
+    let mut file = File::create(CLUSTER_CONFIG_PATH)
+        .context(Resources::Clear, "Failed to create yaml file")?;
+
+    file.write_all(yaml.unwrap().as_bytes()).context(
+        Resources::Clear,
+        format!(
+            "Unable to write eksctl configuration to '{}'",
+            CLUSTER_CONFIG_PATH
+        ),
+    )?;
+    info!("YAML file has been created at /local/cluster_config.yaml");
+
+    Ok(())
 }
 
 impl ClusterConfig {
@@ -143,15 +262,6 @@ impl ClusterConfig {
                         "Unable to convert decoded config to string.",
                     )?)
                     .context(Resources::Clear, "Unable to serialize eksctl config.")?;
-
-                let config_path = Path::new(TEST_CLUSTER_CONFIG_PATH);
-                write(config_path, decoded_config).context(
-                    Resources::Clear,
-                    format!(
-                        "Unable to write eksctl configuration to '{}'",
-                        config_path.display()
-                    ),
-                )?;
 
                 let (cluster_name, region) = config
                     .get("metadata")
@@ -193,9 +303,9 @@ impl ClusterConfig {
         Ok(config)
     }
 
-    /// Create a cluster with the given config.
+    // Create a cluster with the given config.
     pub fn create_cluster(&self) -> ProviderResult<()> {
-        let status = match self {
+        let cluster_config_path = match self {
             Self::Args {
                 cluster_name,
                 region,
@@ -206,41 +316,32 @@ impl ClusterConfig {
                     .as_ref()
                     .map(|version| version.major_minor_without_v())
                     .unwrap_or_else(|| DEFAULT_VERSION.to_string());
-                trace!("Calling eksctl create cluster");
-                let status = Command::new("eksctl")
-                    .args([
-                        "create",
-                        "cluster",
-                        "-r",
-                        region,
-                        "--zones",
-                        &zones.clone().unwrap_or_default().join(","),
-                        "--version",
-                        &version_arg,
-                        "-n",
-                        cluster_name,
-                        "--nodes",
-                        "0",
-                        "--managed=false",
-                    ])
-                    .status()
-                    .context(Resources::Clear, "Failed create cluster")?;
-                trace!("eksctl create cluster has completed");
-                status
+
+                create_yaml(cluster_name, region, &version_arg, zones)?;
+                info!(
+                    "assigned create cluster yaml file path is {}",
+                    CLUSTER_CONFIG_PATH
+                );
+                CLUSTER_CONFIG_PATH
             }
             Self::ConfigPath {
                 cluster_name: _,
                 region: _,
             } => {
-                trace!("Calling eksctl create cluster with config file");
-                let status = Command::new("eksctl")
-                    .args(["create", "cluster", "-f", TEST_CLUSTER_CONFIG_PATH])
-                    .status()
-                    .context(Resources::Clear, "Failed create cluster")?;
-                trace!("eksctl create cluster has completed");
-                status
+                info!(
+                    "assigned create cluster yaml file path is {}",
+                    TEST_CLUSTER_CONFIG_PATH
+                );
+                TEST_CLUSTER_CONFIG_PATH
             }
         };
+
+        trace!("Calling eksctl create cluster with config file");
+        let status = Command::new("eksctl")
+            .args(["create", "cluster", "-f", cluster_config_path])
+            .status()
+            .context(Resources::Clear, "Failed create cluster")?;
+        trace!("eksctl create cluster has completed");
         if !status.success() {
             return Err(ProviderError::new_with_context(
                 Resources::Clear,


bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -20,16 +20,19 @@ use resource_agent::provider::{
 use serde::{Deserialize, Serialize};
```
gthao313, line 20: I think we don't need 4 commits here, and can you squash them to one? https://www.gitkraken.com/learn/git/best-practices/git-commit-message

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +133,103 @@ enum ClusterConfig {
...
+        .context(Resources::Clear, "Failed to create yaml file")?;
+
+    file.write_all(yaml.as_bytes()).context(
+        Resources::Clear,
+        format!(
+            "Unable to write eksctl configuration to '{}'",
+            CLUSTER_CONFIG_PATH
+        ),
+    )?;
+    info!("YAML file has been created at /local/ipv6_config.yaml");
```
gthao313: It could be `/local/ipv4_config.yaml` as well.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +133,103 @@ enum ClusterConfig {
...
+#[derive(Serialize)]
+struct Addon {
+    name: String,
+    version: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct IAMConfig {
+    with_o_i_d_c: bool,
```
gthao313: What does this variant name mean?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -194,38 +286,27 @@ impl ClusterConfig {
...
-                zones,
+                zones: _,
             } => {
                 let version_arg = version
                     .as_ref()
                     .map(|version| version.major_minor_without_v())
                     .unwrap_or_else(|| DEFAULT_VERSION.to_string());
-                trace!("Calling eksctl create cluster");
+
+                let _ = create_yaml(cluster_name, region, &version_arg);
```
gthao313: NIT
```suggestion
                create_yaml(cluster_name, region, &version_arg);
```

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -194,38 +286,27 @@ impl ClusterConfig {
...
-                        &zones.clone().unwrap_or_default().join(","),
-                        "--version",
-                        &version_arg,
-                        "-n",
-                        cluster_name,
-                        "--nodes",
-                        "0",
-                        "--managed=false",
-                    ])
+                    .args(["create", "cluster", "-f", CLUSTER_CONFIG_PATH])
```
gthao313: "There appears to be overlap between `Self::ConfigPath` and `Self::Args`, as both methods currently create clusters using YAML files. The primary distinction lies in their approach to file handling:

Self::Args generates a YAML file internally.
Self::ConfigPath directly utilizes a pre-existing file provided by the user.
Alternatively, I think we can do 
```
let CLUSTER_CONFIG_PATH = match self {
            Self::Args {
                cluster_name,
                region,
                version,
                zones,
            } => {
                create_yaml(cluster_name, region, &version_arg);
                CLUSTER_CONFIG_PATH
            }
            Self::ConfigPath {
                cluster_name: _,
                region: _,
            } => {
                TEST_CLUSTER_CONFIG_PATH
            }
    };

trace!("Calling eksctl create cluster with config file");
let status = Command::new("eksctl")
    .args(["create", "cluster", "-f", CLUSTER_CONFIG_PATH])
    .status()
    .context(Resources::Clear, "Failed create cluster")?;
trace!("eksctl create cluster has completed");

if !status.success() {
    return Err(ProviderError::new_with_context(
        Resources::Clear,
        format!("Failed create cluster with status code {}", status),
    ));
}
```

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -20,16 +20,19 @@ use resource_agent::provider::{
...
 
+use std::fs::File;
+use std::io::Write;
+
 /// The default region for the cluster.
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
 const DEFAULT_VERSION: &str = "1.24";
 const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
+const CLUSTER_CONFIG_PATH: &str = "/local/ipv6_config.yaml";
```
gthao313: Can we just do `/local/cluster_config.yaml`? It could use for ipv4 as well.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: squashed

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: withOIDC: true

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +133,103 @@ enum ClusterConfig {
...
+}
+
+#[derive(Serialize)]
+struct ManagedNodeGroup {
+    name: String,
+}
+
+fn create_yaml(cluster_name: &str, region: &str, version: &str) -> ProviderResult<()> {
+    let ip_family = if cluster_name.ends_with("ipv6") {
+        "IPv6"
```
gthao313: NIT: Can you make those values as mutable variables. Thanks
like
```
const IPV6:  &str = "IPv6";
```

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +133,103 @@ enum ClusterConfig {
...
+        .context(Resources::Clear, "Failed to create yaml file")?;
+
+    file.write_all(yaml.as_bytes()).context(
+        Resources::Clear,
+        format!(
+            "Unable to write eksctl configuration to '{}'",
+            CLUSTER_CONFIG_PATH
+        ),
+    )?;
+    info!("YAML file has been created at /local/ipv6_config.yaml");
```
gthao313: I think this hasn't been changes. It can be `/local/ipv4_config.yaml` as well.

gthao313:
LGTM. Can you fix the checks by following the error messages? thanks!

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -116,6 +121,7 @@ impl AwsClients {
     }
 }
 
+#[allow(dead_code)]
```
yeazelm: What dead code are we allowing?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
...
+    withOIDC: bool,
+}
+
+#[derive(Serialize)]
+struct ManagedNodeGroup {
+    name: String,
+}
+
+fn create_yaml(cluster_name: &str, region: &str, version: &str) -> ProviderResult<()> {
+    let ip_family = if cluster_name.ends_with("ipv6") {
```
yeazelm, line 195: This would probably be better solved with an `Enum` since we know there are two valid values and only two valid values.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
     },
 }
 
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Config {
```
yeazelm, line 147: Can we get doc strings for all these structs? I think this is specifically an `eksctl` configuration struct, but a docstring would help. I think `EksctlConfig` might be a better name as well to signal we don't own this format, its intended to be used specifically as input to `eksctl`.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
...
+struct Metadata {
+    name: String,
+    region: String,
+    version: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct KubernetesNetworkConfig {
+    ip_family: String,
```
yeazelm, line 168: If you made this type an Enum, you could use the type system to protect your values against unknown/invalid ones.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
...
+                version: "latest".to_string(),
+            },
+        ],
+        iam: IAMConfig { withOIDC: true },
+        managed_node_groups: vec![ManagedNodeGroup {
+            name: "mng-1".to_string(),
+        }],
+    };
+
+    let yaml = serde_yaml::to_string(&cluster).expect("Failed to serialize YAML");
```
yeazelm: Can we use `.context()` here to stay consistent with the rest of the codebase on error handling?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -20,16 +20,21 @@ use resource_agent::provider::{
...
 use std::process::Command;
 use testsys_model::{Configuration, SecretName};
 
+use std::fs::File;
+use std::io::Write;
+
 /// The default region for the cluster.
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
 const DEFAULT_VERSION: &str = "1.24";
```
yeazelm, line 33: I know this isn't in your diff but we should consider upgrading our default k8s to a newer version. Not needed in this PR but raising here to remind us to get an issue cut to track moving to a newer k8s for default.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
...
+        ],
+        iam: IAMConfig { withOIDC: true },
+        managed_node_groups: vec![ManagedNodeGroup {
+            name: "mng-1".to_string(),
+        }],
+    };
+
+    let yaml = serde_yaml::to_string(&cluster).expect("Failed to serialize YAML");
+
+    let mut file = File::create(CLUSTER_CONFIG_PATH)
```
yeazelm, line 236: Do repeated runs use the same path? Will `create()` handle running this code a second time? What if this file wasn't put here by the code? Does this break any abilities to run multiple runs of testsys at the same time? Is there a reason we want this around after the fact or could we sidestep these concerns by using a temporary file location for the run?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -129,6 +135,103 @@ enum ClusterConfig {
...
+        .context(Resources::Clear, "Failed to create yaml file")?;
+
+    file.write_all(yaml.as_bytes()).context(
+        Resources::Clear,
+        format!(
+            "Unable to write eksctl configuration to '{}'",
+            CLUSTER_CONFIG_PATH
+        ),
+    )?;
+    info!("YAML file has been created at /local/cluster_config.yaml");
```
yeazelm, line 246: nit: you use the `CLUSTER_CONFIG_PATH` everywhere else, please use it in the `info!()` to avoid bitrot as things are refactored.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -20,16 +20,21 @@ use resource_agent::provider::{
...
 
+use std::fs::File;
+use std::io::Write;
+
 /// The default region for the cluster.
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
 const DEFAULT_VERSION: &str = "1.24";
 const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
+const CLUSTER_CONFIG_PATH: &str = "/local/cluster_config.yaml";
```
yeazelm, line 35: Why did you create your own cluster_config.yaml file here? Do we need both this and `TEST_CLUSTER_CONFIG_PATH`?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -194,53 +288,35 @@ impl ClusterConfig {
     }
 
     /// Create a cluster with the given config.
+    #[allow(clippy::let_and_return)]
```
yeazelm: Why did you add this? I think we want to fix the code, not disable the lint.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -194,53 +288,35 @@ impl ClusterConfig {
...
     pub fn create_cluster(&self) -> ProviderResult<()> {
-        let status = match self {
+        let cluster_config_path = match self {
             Self::Args {
                 cluster_name,
                 region,
                 version,
-                zones,
+                zones: _,
```
yeazelm: I don't follow the change to understand why you needed to change this? Are we no longer handling zones? Won't this break code where we have to be very clear which zone we are launching in when there is capacity only in a certain zone?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -144,15 +247,6 @@ impl ClusterConfig {
                     )?)
                     .context(Resources::Clear, "Unable to serialize eksctl config.")?;
```
yeazelm, line 265: It seems like you just broke the ability to pass in a file for `ClusterConfig`? Can you explain why you removed this?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -194,53 +288,35 @@ impl ClusterConfig {
...
-                    .args(["create", "cluster", "-f", TEST_CLUSTER_CONFIG_PATH])
-                    .status()
-                    .context(Resources::Clear, "Failed create cluster")?;
-                trace!("eksctl create cluster has completed");
-                status
-            }
+            } => TEST_CLUSTER_CONFIG_PATH,
         };
+
+        trace!("Calling eksctl create cluster with config file");
```
yeazelm, line 339: nit: We probably want to indicate *which* configuration file we are calling with. This could be very confusing if its using the one you generate vs one passed in and this log line wouldn't help the user discern which one is happening.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: I didn't change this part of the code, used to add some code here later on I recovery to it original state.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -128,6 +130,123 @@ enum ClusterConfig {
...
+enum IPFamily {
+    IPv4,
+    IPv6,
+}
+
+fn get_family_string(family: IPFamily) -> &'static str {
+    match family {
+        IPFamily::IPv4 => "IPv4",
+        IPFamily::IPv6 => "IPv6",
+    }
```
rpkelly, line 142: Either use `strum_macros` to do this for you, or:

```
impl Display for IPFamily {
    fn fmt(&self, f: &mut Formatter) -> Result {
        match self {
            IPFamily::IPv4 => write!("IPv4"),
            IPFamily::IPv6 => write!("IPv6")
        }
    }
}
````

Then  `IPFamily::IPv4.to_string()` should work, which means that any where that would automatically call `to_string` will.

File: bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs (hunk 17 above)
Comment (rpkelly) [reply bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs:35]: For posterity's sake can you please provide the answer to this question here, and potentially leave a comment in the code to differentiate these configs?

File: bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs (hunk 11 above)
Comment (rpkelly) [reply bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs:147]: +1 on the docstring

File: bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs [hunk 23]
```diff
@@ -128,6 +130,123 @@ enum ClusterConfig {
...
+struct Metadata {
+    name: String,
+    region: String,
+    version: String,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct KubernetesNetworkConfig {
+    ip_family: String,
```
rpkelly, line 168: As long as your enum is given `#[derive(Serialize)]` serde should treat it properly (even if you don't implement `to_string()` as above. If we don't set this to enum, we don't get the benefits of type checking that the enum is trying to grant us.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -128,6 +130,123 @@ enum ClusterConfig {
...
+}
+
+#[allow(clippy::unwrap_or_default)]
+fn create_yaml(
+    cluster_name: &str,
+    region: &str,
+    version: &str,
+    zones: &Option<Vec<String>>,
+) -> ProviderResult<()> {
+    let ip_family = if cluster_name.ends_with("ipv6") {
```
rpkelly, line 195: don't set this as a string, set it as an enum to get the type checking.

PR #974, opened 2025-02-20
eks-resource-agent: default ManagedNodeGroup desiredCapacity to 0

gthao313:
**Issue number:**

**Description of changes:**
```
eks-resource-agent: default ManagedNodeGroup desiredCapacity to 0
```
During testing, we should avoid using AL nodes from ManagedNodeGroup. These nodes cause Karpenter tests to fail because Karpenter attempts to taint them without having the necessary permissions. This results in Karpenter resource creation hanging until it times out.

**Testing done:**

- [x] Deploy the EKS cluster using the modified resource agent image. Then verify that the ManagedNodeGroup's desired capacity is defaulting to 0
Screenshot 2025-02-20 at 3 44 32 PM

- [x] Run Karpenter test

```
x86-64-aws-k8s-128-quick                          Test                    passed                                     5                     0                   7391   f818b28c               2025-02-20T23:32:08Z
 x86-64-aws-k8s-129-quick                          Test                    passed                                     5                     0                   7410   f818b28c               2025-02-20T23:37:22Z
```

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -31,9 +31,13 @@ use strum_macros::EnumString;
 /// The default region for the cluster.
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
-const DEFAULT_VERSION: &str = "1.24";
+const DEFAULT_VERSION: &str = "1.32";
 const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
 const CLUSTER_CONFIG_PATH: &str = "/local/cluster_config.yaml";
+/// The default cluster managed node group values.
+const MNG_MIN_SIZE: i32 = 0;
+const MNG_MAX_SIZE: i32 = 2;
+const MNG_DESIRED_CAPACITY: i32 = 0;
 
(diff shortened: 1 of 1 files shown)


zongzhidanielhu (approved):

mgsharm (approved):

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -31,9 +31,13 @@ use strum_macros::EnumString;
...
 const DEFAULT_REGION: &str = "us-west-2";
 /// The default cluster version.
-const DEFAULT_VERSION: &str = "1.24";
+const DEFAULT_VERSION: &str = "1.32";
 const TEST_CLUSTER_CONFIG_PATH: &str = "/local/eksctl_config.yaml";
 const CLUSTER_CONFIG_PATH: &str = "/local/cluster_config.yaml";
+/// The default cluster managed node group values.
+const MNG_MIN_SIZE: i32 = 0;
+const MNG_MAX_SIZE: i32 = 2;
+const MNG_DESIRED_CAPACITY: i32 = 0;
```
yeazelm, line 40: If we are setting this to 0, why are we creating it at all? Would it be better to fix the permissions error for adding taints than to set this to 0?

yeazelm (approved):

PR #984, opened 2025-03-17
feat: add single zone configuration

zongzhidanielhu:
**Issue number:**

**Description of changes:**
Added feature allows configuring a specific zone in Test.toml, so that  EC2 instance(s) can be launched in this specific zone. Usually, two zones must specified, but this feature allows one specific single zone for EC2 instance to launch into.

**Testing done:**
5 rounds same testings have been launched to make sure us-west-2b is not randomly picked up: 
Input Test.toml: 
Screenshot 2025-03-17 at 11 23 30 AM

After EC2 instance launched: 
[ (eksctl-x86-64-aws-k8s-129-nvidia-cluster/SubnetPublicUSWEST2B)]

Invalid input tests: 
input: 
Screenshot 2025-03-17 at 1 49 13 PM

output: 
Screenshot 2025-03-17 at 1 49 54 PM

**Terms of contribution:**

### DIFF
--- a/bottlerocket/types/src/agent_config.rs
+++ b/bottlerocket/types/src/agent_config.rs
@@ -296,6 +296,9 @@ pub struct Ec2Config {
     /// The device mappings used for EC2 resource provisioning
     #[serde(default)]
     pub device_mappings: Option<Vec<BlockDeviceMappingConfig>>,
+
+     /// The target availibity zones
+     pub input_availability_zones: Vec<String>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize, Default)]


bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -235,8 +235,14 @@ where
     for instance_type in instance_types {
         info!("Using instance type '{}'", instance_type);
 
+        let mut flex_subnet_ids = spec.clone().configuration.subnet_ids;
```
gthao313: Should we do something like this?
```suggestion
        let mut flex_subnet_ids = spec.configuration.subnet_ids.clone();
```

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
```diff
@@ -235,8 +235,14 @@ where
...
+
+        if flex_subnet_ids.len() <= 2 {
+            flex_subnet_ids.remove(0);
```
gthao313: I would explain why we should remove the first position value here and why this would be triggered under two subnet ids.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -142,6 +143,17 @@ enum IPFamily {
...
+#[derive(Debug, EnumString, AsRefStr)]
+#[strum(serialize_all = "kebab-case")]
+enum AvailabilityZones {
+    #[strum(serialize = "us-west-2a")]
+    UsWest2a,
+    #[strum(serialize = "us-west-2b")]
+    UsWest2b,
+    #[strum(serialize = "us-west-2c")]
+    UsWest2c,
+}
```
gthao313: If we miss `UsWest2d` here?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -239,10 +251,23 @@ struct ManagedNodeGroup {
     min_size: i32,
     /// The maximum number of nodes in the managed node group.
     max_size: i32,
-    // The desired number of nodes in the managed node group.
+    /// The desired number of nodes in the managed node group.
     desired_capacity: i32,
 }
 
+fn get_remaining_zones(exclude: String) -> Vec<AvailabilityZones> {
```
gthao313: I would rename `exclude` to like `zone_to_exclude `

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -256,6 +281,24 @@ fn create_yaml(
         IPFamily::IPv4
     };
 
+    let mut eks_zones: Vec<String> = zones.clone().unwrap();
```
gthao313: I would recommend to not use `.unwrap()` directly here as it will panic with a generic error message if `zones` is None. Instead, use .context() or similar error handling to provide a more meaningful error message
```suggestion
    let mut eks_zones: Vec<String> = zones.clone().context("Failed to get availability zones for EKS cluster")?;
```

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -256,6 +281,24 @@ fn create_yaml(
...
+
+    if zones.is_some() && zones.iter().len() < 2 {
+        let zones_upwrap = zones.clone().unwrap().first().unwrap().to_string();
```
gthao313: Same as above. : )

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -256,6 +281,24 @@ fn create_yaml(
...
+        if AvailabilityZones::from_str(&zones_upwrap).is_err() {
```
gthao313: I feel like we should run the zone input value validation at twoliter. Twoliter does the job to read `test.toml` file and pass it to testsys. so I think twoliter should own the job to validate it as well.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -256,6 +281,24 @@ fn create_yaml(
...
+                Resources::Clear,
+                format!("Input zone of {} in Test.toml is invalid", zones_upwrap),
+            ));
+        }
+        let remain_zone = get_remaining_zones(zones_upwrap)
+            .first()
+            .unwrap()
+            .as_ref()
+            .to_string();
+        eks_zones.insert(0, remain_zone);
```
gthao313: I think I don't fully understand this method. The second position value will always be `2b` ?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -142,6 +143,17 @@ enum IPFamily {
     IPv4,
 }
 
+#[derive(Debug, EnumString, AsRefStr)]
+#[strum(serialize_all = "kebab-case")]
+enum AvailabilityZones {
+    #[strum(serialize = "us-west-2a")]
+    UsWest2a,
```
yeazelm: I'm not sure if we want to hard code this. Is there a way we can get the Availability Zones from the ec2 client at runtime?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
gthao313: I think I understand the current implementation (although I haven't seen the twoliter logic directly). The system randomly adds an extra availability zone to the EKS cluster zone list if only one zone is provided by Test.toml and inserts its subnet ID at the start of the subnet list. When launching EC2 instances, it removes this first subnet ID and uses the remaining ones. I think this isn't the ideal solution that we want.

Instead, we should use the AWS describe-subnets API to identify the exact subnet IDs matching our target availability zones from Test.toml, then pass these specific subnet IDs to ec2 instance launch client. This would allow the EC2 agent to launch instances only in our intended subnets, providing a cleaner and more controlled solution.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
gthao313: It appears the test is flagging UsWest2d as an invalid input, which doesn't seem correct. Why should we consider this availability zone invalid when it's a legitimate zone in AWS US-West-2 region?

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: I disagree of using AWS describe-subnets, this is just add another layer of unnecessary calling of other API, the hard part of this problem is when have 2 subnets, you don't know which on is which, here I just used the most simple way, the Vec index to identify the target subnet without calling anything API overhead, and remove the others. You don't have to describe the subnets cause you already known the index[1] is the target zone subnet id. 

By adding up the AWS describe-subnets, I need to setup the awscli, and do a API call, get the response, then from the response, find the field need, compare what I did which is much simple and clean, this just too much overhead.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
zongzhidanielhu: sure, Will do

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: There are 3 valid zones a, b and c. but in Test.toml you basically can enter whatever you want to. This check just make sure input is valid, if not log out info in testsys. eksctl will through a error too if the input zone is invalid, but a concise code will do a check too, I would rather check it before the eksctl through.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: either way will works, but checking here is much easily due to the core logic is in the cluster creation, and the key purpose of give a input zone is for create a cluster with the instance we want to.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
zongzhidanielhu: This is the hard part of this issue, everything has to be done before the ec2 instances created, that's why I can not use ec2 client due to by then the instance(s) already created which is no way we can manipulate it.  It is kind of not hardcode due to it is only has three zones, a, b, c, and it is just for define the valid zone boundary. This will give us more flexible of when input is c, you know you will add a or b to create cluster, when input is b, you know you will check a, c, not e, f, g cause the valid zone boundary already defined.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
yeazelm: > By adding up the AWS describe-subnets, I need to setup the awscli, and do a API call, get the response, then from the response, find the field need, compare what I did which is much simple and clean, this just too much overhead.

You should use the AWS Rust SDK to make these calls and I agree with @gthao313, we should do this. It is not too much overhead, you have to make this code functional in more than just this one situation, and these describes and dealing with dynamic changes is part of dealing with making this code re-useable/flexible.

> You don't have to describe the subnets cause you already known the index[1] is the target zone subnet id.

You have hard coded the AZs above which is why I commented there. This code will only ever work while there are 3 AZs and only in us-west-2. Using describe calls with the AWS SDK would allow this code to run other places than us-west-2 and still retain the same functionality.

bottlerocket/agents/src/bin/ec2-resource-agent/ec2_provider.rs
yeazelm: What if someone provides `["us-west-2b", "us-west-2c"]`, wouldn't this now fail? I do not believe we want fixed offsets like this. We should look for the value we want and use that. I don't think it is safe to assume the input is sorted or in the order we expect.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
yeazelm: I think we need to go back to the approach and see if we need to create an EC2 client before the rest of this because this is a blocking issue, we can't merge hard-coding us-west-2 AZs as the only place this works.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
yeazelm: Your account might only have 3 but there can be four https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-availability-zones.html

yeazelm (changes requested):

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -239,10 +251,23 @@ struct ManagedNodeGroup {
     min_size: i32,
     /// The maximum number of nodes in the managed node group.
     max_size: i32,
-    // The desired number of nodes in the managed node group.
+    /// The desired number of nodes in the managed node group.
     desired_capacity: i32,
 }
 
+fn zone_to_exclude(exclude: String) -> Vec<AvailabilityZones> {
```
yeazelm: This feels like it should be swapped out for a function using the EC2 client to get the available availability zones for that account in that region. 

nit: the name is not really helpful. You have named it what you are using it for, not what it does. This function returns the availability zones in us-west-2. If we must keep this function (which my above comment asks not to), then it should be named something more specific to its function, not what you want to use it for like `all_us_west_2_zones()` or something like that.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
```diff
@@ -256,6 +281,35 @@ fn create_yaml(
...
+
+        if AvailabilityZones::from_str(zone_input).is_err() {
+            return Err(ProviderError::new_with_context(
+                Resources::Clear,
+                format!("Input zone of {} in Test.toml is invalid", zone_input),
+            ));
+        }
+        let remain_zone = zone_to_exclude(zone_input.to_string())
+            .first()
+            .unwrap()
```
yeazelm: We don't use `unwrap()` for these situations. I like to think of `unwrap()` as a good indicator I might have a flaw in my approach. It makes me ask "why does Rust think this can fail?" and it helps shift my focus to how to avoid it being possible to fail. In this case, you likely need to handle the error since you are hard coding the output from this function, you should be able to enforce that it won't fail. But if this was output from a client to EC2, then you need to properly handle if the output of this function doesn't match your expectations of there being a return from `first()`.

bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
gthao313: I agree that we shouldn't hardcode since it would be broken if we pass the zone at other region. It already provides the region that cluster is running at and setup aws config based on the region, so we can programmatically list the available zones by using the ec2_client.

The issue here is to provide extra available zones if only one zone is provided via Test.toml, so we list all zones and randomly give one (not the input one) to eksctl. 

This would help https://docs.rs/aws-sdk-ec2/latest/aws_sdk_ec2/struct.Client.html#method.describe_availability_zones

PR #994, opened 2025-05-30
Add kubernetes 1.33 support

jpculp:
**Description of changes:**

Sets the amiFamily in the initial managed node group to Bottlerocket, as AmazonLinux2 is not a valid amiFamily for k8s 1.33.

**Testing done:**

```bash
 x86-64-aws-k8s-133-quick                    Test                        passed                                             5                         0                       6730   97fcd255                   2025-06-05T00:46:20Z
 x86-64-aws-k8s-133                          Resource                    completed                                                                                                   97fcd255                   2025-06-05T00:44:04Z
```
(thanks, @zongzhidanielhu!)

**Terms of contribution:**

### DIFF
--- a/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
+++ b/bottlerocket/agents/src/bin/eks-resource-agent/eks_provider.rs
@@ -142,6 +142,11 @@ enum IPFamily {
     IPv4,
 }
 
+#[derive(Serialize, Debug, EnumString)]
+enum AmiFamily {
+    Bottlerocket,
+}
+
 /// Configuration for setting up an EKS cluster using eksctl yaml file.
 ///
 /// # Fields:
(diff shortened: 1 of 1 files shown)


yeazelm (approved):
LTGM assuming the testing works well. I think its a nice improvement to avoid creating the node groups but I wonder if the clusters will have health issues without any node to run things like CoreDNS and other cluster-level daemonsets?

jpculp:
> I wonder if the clusters will have health issues without any node to run things like CoreDNS and other cluster-level daemonsets?

I haven't seen that be the case in my personal EKS usage.

jpculp:
Due to the following error, I'm removing the commit where we skip the nodegroup creation for now:
```bash
[2025-06-05T00:16:25Z ERROR resource_agent::agent] Resource action failed: Provider error: An error left resources behind that can be destroyed, Could not find nodegroup cloudformation stack for cluster
```

zongzhidanielhu (approved):

yeazelm (approved):

PR #999, opened 2025-06-12
workload: add egl check to nvidia-smoke

jpculp:
**Description of changes:**

Add EGL check to the list of tests to run.

**Testing done:**

Ran on two Bottlerocket AMIs, one built using a kernel kit without `libnvidia-gpucomp.so` and one with.
The EGL test passed on the latter, and failed on the former.

**Terms of contribution:**

### DIFF
--- a/bottlerocket/tests/workload/nvidia-smoke/Dockerfile
+++ b/bottlerocket/tests/workload/nvidia-smoke/Dockerfile
@@ -3,12 +3,17 @@
 # the larger image, then copy them to the lightweight image we use to run.
 FROM nvidia/cuda:12.9.0-devel-amzn2023 AS builder
 
-RUN dnf makecache \
-  && dnf -y install \
-    cmake \
-    git-core \
-  && dnf clean all \
-  && rm -fr /var/cache/dnf
+RUN dnf makecache && \
+  dnf install -y \
+  cmake \
+  g++ \
+  git-core \
+  libX11-devel \
+  python-devel \
+  python-pip \
+  && \
+  dnf clean all && \
+  rm -rf /var/cache/dnf
 
 # Clone the samples, pinned to a version we know should work
 RUN git clone --branch v12.9 --depth 1 https://github.com/NVIDIA/cuda-samples.git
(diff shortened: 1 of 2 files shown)


bottlerocket/tests/workload/nvidia-smoke/Dockerfile
```diff
@@ -31,17 +36,24 @@ RUN cd /cuda-samples/Samples/3_CUDA_Features/warpAggregatedAtomicsCG/ && cmake .
...
 RUN dnf makecache && \
-    dnf install -y \
-      gzip \
-      tar \
-    && \
-    dnf clean all && \
-    rm -rf /var/cache/dnf
+  dnf install -y \
+  gzip \
+  libglvnd-devel \
```
arnaldo2792, line 49: Are these headers needed, or do you only need `libglvnd` at runtime?

bottlerocket/tests/workload/nvidia-smoke/Dockerfile
```diff
@@ -3,12 +3,17 @@
...
-    git-core \
-  && dnf clean all \
-  && rm -fr /var/cache/dnf
+RUN dnf makecache && \
+  dnf install -y \
+  cmake \
+  g++ \
+  git-core \
+  libX11-devel \
+  python-devel \
```
arnaldo2792, line 12: Are these needed because `moderngl` requires them at install time?

arnaldo2792 (approved):

bottlerocket/tests/workload/nvidia-smoke/Dockerfile
jpculp: Correct.

bottlerocket/tests/workload/nvidia-smoke/Dockerfile
jpculp: This provides `libEGL.so` at runtime.

jpculp: I think `libglvnd-egl` also works since it provides `libEGL.so.1`, but it only ends up saving us 9mb.

yeazelm (approved):

